PR #2, opened 2015-04-05
Simplify File creation with splitext

mbauman:


mbauman:
Ah, thanks, I didn't catch the lowercasing.

SimonDanisch:
We might need a more sophisticated process for this anyways. Some normalized form to handle duplicates and names that map to the same type (e.g. jpg, jpeg, JPG)

PR #10, opened 2015-04-20
Refactor and one more test

brk00:
Following the structure from most julia packages, FileIO.jl is just the module definition with imports, exports, etc. The functionality is defined on `core.jl` that is `include`d into the module definition.

I also added some comments and changed the tests for a `.txt` file that I created. I added a test for an inexistent file.

### DIFF
--- a/src/FileIO.jl
+++ b/src/FileIO.jl
@@ -1,46 +1,19 @@
 module FileIO
-import Base: read
-import Base: write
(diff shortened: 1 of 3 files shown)


SimonDanisch:
Thanks this looks good!
Writing exports and imports like you do is probably better. I was writing them like that mainly for visual reasons and got used to it ;)

PR #15, opened 2015-08-02
Rewrite from scratch

timholy:
This adds magic number support, and tries to address real-world
complications such as:
- The same extension might mean two or more different formats
- The extension might lie about the format
- Some ~~files~~formats don't have magic numbers
- Some magic numbers are found in weird places in files

I'll add some documentation for this, but I may be done for the day.


timholy:
OK, I think this is in good shape now. Probably the most thoroughly-documented module I've ever created, too.

timholy:
Note that reviewing this might be easiest if you first work through the README, which you can do most easily from the front page and selecting this branch.

src/query.jl
```diff
@@ -0,0 +1,257 @@
...
+function magic_cmp(t::Tuple, p::Pair)
+    pt = first(p)
+    lp, lt = length(pt), length(t)
+    lt < lp && return true
+    lt > lp && return false
+    t < pt
+end
+
+
+abstract Formatted{F<:DataFormat}   # A specific file or stream
```
sjkelly: Would `AbstractFormat` work here?

src/query.jl
```diff
@@ -0,0 +1,257 @@
...
+unknown{sym}(::DataFormat{sym}) = unknown(sym)
+unknown(::Any) = false
+
+const ext2sym = Dict{ASCIIString,Union(Symbol,Vector{Symbol})}()
+const magic_list = Array(Pair, 0)    # sorted, see magic_cmp below
+const sym2info = Dict{Symbol,Any}()  # Symbol=>(magic, extension)
+const magic_func = Array(Pair, 0)    # for formats with complex magic #s
+
+@doc """
+`add_format(fmt, magic, extention)` registers a new `DataFormat`.
```
sjkelly: It might be good to mention that extensions are case sensitive.

src/query.jl
timholy: Yes, and that's what I had originally. But somehow `AbstractFormat` sounds to me more like a format specification, whereas these are supposed to represent instances (e.g., particular files).

src/query.jl
timholy: Good point. I'll do that, and at the same time merge all my README tweaks into the original commit.

ssfrr:
This is fantastic, I really like the API, it feels very right.

I'm looking forward to packages not needing their own `MyPkg.open` or `open_myformat` methods.

timholy:
Glad you like it, @ssfrr.

> I'm looking forward to packages not needing their own MyPkg.open or open_myformat methods.

There may still be some of that; for example, are we really going to try to register each of the ~80 file formats supported by ImageMagick? As an alternative, Images might still retain an `imread` that says "use ImageMagick if nothing more specific is suggested."

ssfrr:
OK, maybe I just meant I'm looking forward to AudioIO not needing an `AudioIO.open`. :smile_cat: 

I'll probably end up registering whatever libsndfile supports, and using Wav.jl for .wav files. I'll have to think a bit about what options will make sense.

SimonDanisch:
Looks great!!! I can't express how glad I am that you simply tackled this ;)
It seems to cover all my use cases, so I'll just start using it and see how it goes.

timholy:
Great. I'll merge this, and start playing with it myself. I am sure there will need to be little tweaks made here and there as we try it out in packages.

PR #16, opened 2015-08-03
RFC: documentation on handling tricky cases

timholy:
This documentation PR outlines my thoughts on how to handle some of the inevitable difficulties that will arise. Suggestions welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,12 @@ s = query(io)   # io is a stream
 will return a `File` or `Stream` object that also encodes the detected
 file format.
 
+## Help & Quickstart
+
+You can get an API overview by typing `?FileIO` at the REPL prompt.
+Individual functions have their own help too, e.g., `?query`.
+
+
 ## Adding new formats
 
 You register a new format by calling `add_format(fmt, magic,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,7 +104,23 @@ function save(f::File{format"PNG"}, data)
...
 
-## Help
+## Issues and tricky/interesting cases
 
-You can get an API overview by typing `?FileIO` at the REPL prompt.
-Individual functions have their own help too, e.g., `?add_format`.
+**Issues**:
+1. Package A and package B can both load format FMT. They both define `FileIO.load(file::File{format"FMT"})`, in which case whichever package gets loaded last "wins."
+2. A is better at loading format FMT 80% of the time, but every once in a while A gets partway through the load and realizes it can't handle it. In that case, Package A would like to defer to package B.
+
```
timholy: These two related issues are exemplified by the relationship between `ImageMagick` and `OSXnative` inside `Images`.

README.md
```diff
@@ -98,7 +104,23 @@ function save(f::File{format"PNG"}, data)
...
+**Solution**: A and B should define the non-exported function `load_(file::File{format"FMT"})`, so that users (or A) can call it as `B.load_(file)`. Each package's `load` function should simply call `load_`.
+
+**Issue**:
+Package A is a low-level wrapper of some C library, and its `load` function for FMT returns a raw array. Package B sits on top of A and wraps the array in some fancy container. When the user says `load(file)` and `file` has format FMT, which return type should be chosen?
+
```
timholy: This issue is exemplified in the relationship between `ImageMagick` and `Images`.

README.md
```diff
@@ -98,7 +104,23 @@ function save(f::File{format"PNG"}, data)
 end
 ```
 
-## Help
+## Issues and tricky/interesting cases
 
-You can get an API overview by typing `?FileIO` at the REPL prompt.
-Individual functions have their own help too, e.g., `?add_format`.
+**Issues**:
+1. Package A and package B can both load format FMT. They both define `FileIO.load(file::File{format"FMT"})`, in which case whichever package gets loaded last "wins."
```
sjkelly: I feel like this should be generalized since it is the behavior of all method signatures in Julia.

README.md
```diff
@@ -98,7 +104,23 @@ function save(f::File{format"PNG"}, data)
...
+2. A is better at loading format FMT 80% of the time, but every once in a while A gets partway through the load and realizes it can't handle it. In that case, Package A would like to defer to package B.
+
+**Solution**: A and B should define the non-exported function `load_(file::File{format"FMT"})`, so that users (or A) can call it as `B.load_(file)`. Each package's `load` function should simply call `load_`.
+
+**Issue**:
+Package A is a low-level wrapper of some C library, and its `load` function for FMT returns a raw array. Package B sits on top of A and wraps the array in some fancy container. When the user says `load(file)` and `file` has format FMT, which return type should be chosen?
+
+**Solution**: It seems a fair assumption that if B is loaded, users probably want its extra features. However, without B loaded, it seems reasonable to choose A. Consequently, FileIO should register A (the lowest-level package) as the loader for FMT. When users say `using B`, assuming that B has the lines
```
sjkelly: I agree this is definitely the typical behavior people would expect when `using` higher level packages. Maybe this could be solved with a 'reverse call' syntax, i.e.
```
function FileIO.load(f::FileIO::Formatted, t::Type{T}, opts...)
    T(load(f, opts...))
end
```

timholy:
I think this is outdated now.

PR #25, opened 2015-08-31
some improvements

SimonDanisch:
Fixed a bug, added precompile and some more mesh formats

### DIFF
--- a/src/query.jl
+++ b/src/query.jl
@@ -148,7 +148,7 @@ file.""" ->
 immutable File{F<:DataFormat} <: Formatted{F}
     filename::UTF8String
 end
-File(fmt::DataFormat, filename) = File{fmt}(filename)
+File{sym}(fmt::Type{DataFormat{sym}}, filename) = File{fmt}(filename)
 
 @doc """
 `filename(file)` returns the filename associated with `File` `file`.
@@ -256,7 +256,7 @@ function query(filename::AbstractString)
     if haskey(ext2sym, ext)
         sym = ext2sym[ext]
         len = lenmagic(sym)
-        if length(len) == 1 && (all(x->x==0, len) || !isfile(filename)) # we only found one candidate and there is no magic bytes, or no file, trust the extension
+        if length(len) == 1 && (((all(x->x==0, len) || !isfile(filename))) || !isfile(filename)) # we only found one candidate and there is no magic bytes, or no file, trust the extension
             return File{DataFormat{sym}}(filename)
         elseif !isfile(filename) && length(len) > 1
             error("no file for check of magic bytes and multiple extensions possible: $sym")
@@ -265,6 +265,7 @@ function query(filename::AbstractString)
             error("Some formats with extension ", ext, " have no magic bytes; use `File{format\"FMT\"}(filename)` to resolve the ambiguity.")
         end
     end
+    !isfile(filename) && File{unknown_df}(filename) # (no extension || no magic byte) && no file
     # Otherwise, check the magic bytes
     file!(query(open(filename), filename))
 end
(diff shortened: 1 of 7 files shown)


SimonDanisch:
seem like I need to add a test for a succesfull detection with `detect_stlbinary` to make coverage happy.

src/query.jl
```diff
@@ -240,7 +240,9 @@ For a plain IO object, you can use `skipmagic(io, fmt)`.
 skipmagic{F}(s::Stream{F}) = (skipmagic(stream(s), F); s)
 function skipmagic{sym}(io, fmt::Type{DataFormat{sym}})
     magic, _ = sym2info[sym]
-    seek(io, length(magic))
+    if !isa(magic, Function)
```
timholy: Good catch

src/registry.jl
```diff
@@ -143,3 +143,37 @@ add_saver(format"PLY_BINARY", :MeshIO)
...
+    seekend(io)
+    len = position(io)
+    seekstart(io)
+    len < size_header && return false
+    
+    header = readbytes(io, 80)
+    header[1:6] == b"solid " && return false
+    number_of_triangle_blocks = read(io, Uint32)
+     #1 normal, 3 vertices in Float32 + attrib count, usually 0
+    len != (number_of_triangle_blocks*size_triangleblock)+size_header && return false
```
timholy: Why oh why do people design file formats with magic numbers?? This looks to be as careful as you can be.

timholy:
Looks great! (Yes, a test file would be nice.)

src/registry.jl
```diff
@@ -129,17 +130,59 @@ add_saver(format"WPG", :ImageMagick)
...
+
+function detect_stlascii(io)
+    position(io) != 0 && return false
+    seekend(io)
+    len = position(io)
+    seekstart(io)
+    len < 80 && return false
+    header = readbytes(io, 80) # skip header
+    seekstart(io)
+    return header[1:6] == b"solid " && !detect_stlbinary(io)
```
dreammaker: I think the binary stl detection looks good.  But does the ascii parsing know to rewind the io and reuse the header?  I think it can have useful data, unlike the binary version.

dreammaker: Nevermind.  I see the `seekstart(io)` call now.  Looks good to me!

test/fileio_packages.jl
```diff
@@ -0,0 +1,15 @@
+const all_packages = unique([values(FileIO.sym2loader)..., values(FileIO.sym2saver)...])
+
+for pkg in all_packages
+	if !(pkg in [:ImageMagick, :GLAbstraction, :AndorSIF, :NRRD, :MeshIO]) # temporary solution until they're registered.
```
sjkelly: This seems like a chicken and egg problem.

src/registry.jl
```diff
@@ -129,17 +130,59 @@ add_saver(format"WPG", :ImageMagick)
...
+    skip(io, 80) # skip header
+    number_of_triangle_blocks = read(io, Uint32)
+     #1 normal, 3 vertices in Float32 + attrib count, usually 0
+    len != (number_of_triangle_blocks*size_triangleblock)+size_header && return false
+    skip(io, number_of_triangle_blocks*size_triangleblock-sizeof(Uint16))
+    attrib_byte_count = read(io, Uint16) # read last attrib_byte
+    attrib_byte_count != zero(Uint16) && return false # should be zero as not used
+    eof(io) && return true
+    false
+end
```
dreammaker: I know I commented on the ascii function before, but I actually needed to add a `seekstart(io)` call in the binary detection function to make this work in Meshes.jl.  See the pull request here: JuliaGeometry/Meshes.jl#58.

sjkelly: https://github.com/JuliaIO/FileIO.jl/pull/29

sjkelly:
Merged in #27

PR #27, opened 2015-09-06
Rebase of sd/improvements

sjkelly:
Also try to make CI pass so we can get closer to publishing this.


SimonDanisch:
I don't really know why appveyor isn't picking up this PR, but it seems to work fine for master...

SimonDanisch:
I'm inclined to merge this and tag FileIO. I'm not sure if I forgot anything, but most of the problems should now be in external packages. [Images](https://github.com/timholy/Images.jl/pull/352) will take some more time but if we tag now we can at least advance with Meshes.

src/query.jl
```diff
@@ -254,7 +258,7 @@ function query(filename::AbstractString)
     if haskey(ext2sym, ext)
         sym = ext2sym[ext]
         len = lenmagic(sym)
-        if length(len) == 1 && (all(x->x==0, len) || !isfile(filename)) # we only found one candidate and there is no magic bytes, or no file, trust the extension
+        if length(len) == 1 && (((all(x->x==0, len) || !isfile(filename))) || !isfile(filename)) # we only found one candidate and there is no magic bytes, or no file, trust the extension
```
sjkelly: Parenthesis are not needed here.

src/query.jl
```diff
@@ -313,8 +318,13 @@ seekable(::Any) = false
 
 function iter_eq(A, B)
     length(A) == length(B) || return false
-    for (a,b) in zip(A,B)
-        a == b || return false
+    i,j = 1,1
+    for _=1:length(A)
```
sjkelly: `eachindex` is a faster alternative on 0.4

sjkelly: Nevermind, this is also a 0.3 package.

sjkelly:
@timholy any objections to merging this? I think we are antsy to get MeshIO published. Semver should allow us to tag and if we need to make any API changes it can bump the minor version. Anything that gets registered can come in as a patch version bump.

src/query.jl
```diff
@@ -254,7 +258,7 @@ function query(filename::AbstractString)
     if haskey(ext2sym, ext)
         sym = ext2sym[ext]
         len = lenmagic(sym)
-        if length(len) == 1 && (all(x->x==0, len) || !isfile(filename)) # we only found one candidate and there is no magic bytes, or no file, trust the extension
+        if length(len) == 1 && ((all(x->x==0, len) || !isfile(filename)) || !isfile(filename)) # we only found one candidate and there is no magic bytes, or no file, trust the extension
```
timholy: I'm confused about this line: aren't you effectively doing `isfile(filename)` twice?

src/query.jl
timholy: Though with Compat, `eachindex` is available.

src/registry.jl
```diff
@@ -129,17 +129,59 @@ add_saver(format"WPG", :ImageMagick)
 # Mesh formats
 
 add_format(format"OBJ", (), ".obj")
-add_loader(format"OBJ", :WavefrontObj)
-add_saver(format"OBJ", :WavefrontObj)
+add_loader(format"OBJ", :MeshIO)
```
timholy: Not that it much matters, but originally I had this file organized into "simple cases" (no function needed) at the top, and "complex cases" at the bottom. They are now intermingled; might be best to delete those comments.

timholy:
No objections here, merge when ready.

src/query.jl
SimonDanisch: Uhm, I thought I cleaned that up... Definitely wrong :D

src/registry.jl
SimonDanisch: no, I like it ordered :D The solution here is to make me read more comments...

PR #31, opened 2015-09-18
first take on multiple magic bytes

SimonDanisch:
fun fact: `skipmagic` is surprisingly annoying for multiple magic
bytes... We should probably just save the matched magic byte into the
file format when we query it. Would probably also help the parsers ;)


SimonDanisch:
need to add tests!

PR #37, opened 2015-09-28
Add support for writemime

timholy:
This tries to support `writemime` in a manner similar to the file format API. In this case, the main challenge is handling the "data" argument of `writemime`, whose type may be defined in a package. Consequently, we expect the `add_mime` calls will often be made from external packages.

The main advantage this brings, over "static" implementation of `writemime`, is the ability to select different implementations based on installed packages, etc.

CC @SimonDanisch, @stevengj.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,10 +76,10 @@ This means MeshIO supports loading and saving of the `off` format.
 You can add multiple loaders and specifiers like this:
 ```jl
 add_format(
-    format"BMP", 
+    format"BMP",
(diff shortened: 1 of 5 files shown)


SimonDanisch:
I'll just fix the coverage and this is good to go =)

SimonDanisch:
okay I'm ready for a merge!

timholy:
Thanks for adding the test!

PR #38, opened 2015-09-28
Choose a default for ambiguous formats

timholy:
`save("myfile.ppm", img)` used to generate an error, because there are two formats (`:PPMBinary` and `:PPMText`) defined. Users will rarely care, and in almost all cases you want `:PPMBinary` anyway. So I changed the behavior from "throw an error upon creating a new file when the format is ambiguous" to "choose the first one in the list".

I also provided a new mechanism for forcing the format: `save(format"PPMBinary", "myfile.ppm", img)`. You might have thought that `save(File(format"PPMBinary", "myfile.ppm"), img)` would have been the better choice, but that runs into a problem: unless the (new) Netpbm.jl package has already been manually loaded, it falls back to the "error: no saver defined for ..." method. So we need a method with distinguishable signature, and this seemed to be the obvious one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ notifications:
   email: false
 script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+    - julia -e 'Pkg.clone("https://github.com/JuliaIO/Netpbm.jl.git")'
     - julia -e 'Pkg.clone(pwd()); Pkg.build("FileIO"); Pkg.test("FileIO"; coverage=true)'
 after_success:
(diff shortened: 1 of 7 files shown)


SimonDanisch:
I tried to debug this but seems like it's better if you fix that error.
I guess mapinfo is only defined for `Gray{U8}` ?
But then I ran into a method error with `mybswap(::Gray{U8})`, which is where I decided to stop^^

timholy:
Ah, it's fixed locally but I need to push.

SimonDanisch:
rest seems to work fine!
I'm definitely in favor of your proposal.
Back when we had that discussion I was just happy that there is already a solution, but this is definitely better =)

PR #41, opened 2015-09-30
RFC: Add LoaderError and WriterError exceptions

rsrock:
When throwing these errors in library code, FileIO will
emit an info message and then try the next registered loader / writer.

Simon pointed me in the right direction for this.

Worth a naming check, and perhaps some tests before merging?

### DIFF
--- a/src/FileIO.jl
+++ b/src/FileIO.jl
@@ -36,11 +36,35 @@ export DataFormat,
        stream,
        unknown
 
+import Base.showerror
+
 include("query.jl")
 include("loadsave.jl")
 include("registry.jl")
 
+@doc """
+`LoaderError` should be thrown when loader library code fails, and other libraries should
+be given the chance to recover from the error.  Reports the library name and an error message:
+LoaderError("ImageMagick", "Foo not available")
+""" ->
+type LoaderError <: Exception
+    library::UTF8String
+    msg::UTF8String
+end
+Base.showerror(io::IO, e::LoaderError) = println(io, e.library, " load error: ",
+                                                 msg, "\n  Will try next loader.")
 
+@doc """
+`WriterError` should be thrown when writer library code fails, and other libraries should
+be given the chance to recover from the error.  Reports the library name and an error message:
+WriterError("ImageMagick", "Foo not available")
+""" ->
+type WriterError <: Exception
+    library::UTF8String
+    msg::UTF8String
+end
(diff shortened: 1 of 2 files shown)


rsrock:
Tests pass for me locally on 0.4-rc3.  Note that Travis is currently testing 0.3 and 0.5 here, but not 0.4

src/FileIO.jl
```diff
@@ -36,11 +36,35 @@ export DataFormat,
...
+
 include("query.jl")
 include("loadsave.jl")
 include("registry.jl")
 
+@doc """
+`LoaderError` should be thrown when loader library code fails, and other libraries should
+be given the chance to recover from the error.  Reports the library name and an error message:
+LoaderError("ImageMagick", "Foo not available")
+"""
```
timholy: On 0.3, this line needs to be
```
""" ->
```

timholy:
You might want to modify the .travis.yml file so [this section](https://github.com/JuliaIO/FileIO.jl/blob/35962681fa18fe7d58760f3907bc85ecfc8827ae/.travis.yml#L6-L7) reads
```
 - 0.3
 - 0.4
 - nightly
```
and then push again (once you've fixed that bug on 0.3).

timholy:
How does `LoaderError` get thrown? Note that on nightly, the error is `LoadError` (a type defined in base). Did you forget to push changes to `src/loadsave.jl`?

rsrock:
You'll `throw(LoaderError("QuartzImageIO", "Foo is insufficiently fooey")` when one of your assertions fails, and you want to explicitly hand off to another reader.

Now I'm seeing that `LoadError` here too, after rebasing off the latest here.  I'll try to track that down.

By the way, I'm open to suggestions for better names for `LoaderError` and `WriterError`, especially if they clash somewhat with Base.

SimonDanisch:
I want to add all the other errors and the create a solution for #42.
Should I just merge this and work from there, or keep pushing to your repo?

rsrock:
If you want to merge and go from here, go for it.  I'm trying to figure out the test failure in "multiple libs".  Something changed after 35962681fa, which means that this approach no longer works (or perhaps it never did, but 35962681fa exposed that...)

SimonDanisch:
that's simply because of your rethrow for anything not being a `LoadError`, from what I've seen.

PR #43, opened 2015-10-01
rewrote error handling

SimonDanisch:


SimonDanisch:
next step: use `handle_error()` to determine what to do.

src/loadsave.jl
```diff
@@ -1,37 +1,119 @@
...
 const sym2saver  = Dict{Symbol,Vector{Symbol}}()
-const mimedict   = Dict{Symbol,Vector{Any}}()
+const mimedict   = Dict{Symbol,Vector{@compat(Tuple{DataType, Symbol})}}()
 
-for (appl,fchk,fadd,dct) in (
-        (:applicable_loaders, :check_loader, :add_loader, :sym2loader),
-        (:applicable_savers,  :check_saver,  :add_saver,  :sym2saver))
+function is_installed(pkg::Symbol)
+    isdefined(pkg) && isa(Main.(pkg), Module) && return true # is a module defined in Main scope
+    path = Base.find_in_path(string(pkg)) # hacky way to determine if a Package is installed
```
timholy: Would `return Pkg.installed(pkg) != nothing` be better?

src/loadsave.jl
```diff
@@ -1,37 +1,119 @@
...
+    path == nothing && return false
+    return isfile(path)
+end
+
+function save_import(pkg::Symbol)
```
timholy: `checked_import`? The current name suggests it's for `save`, not for `load`.

src/loadsave.jl
```diff
@@ -1,37 +1,119 @@
...
+- `load(f; options...)` passes keyword arguments on to the loader.
+""" ->
+function load(s::@compat(Union{AbstractString,IO}), args...; options...)
+    q           = query(s)
+    libraries   = applicable_loaders(q)
+    failures    = Any[]
+    for library in libraries
+        try
+            Library = save_import(library)
+            return Library.load(q, args...; options...)
```
timholy: We need to decide once-and-for-all whether we intend to have `load` be overloaded (meaning, if so you should drop the `Library.` in front of `load`). I think it needs to be, particularly for `Stream`s, where users are likely to have to call `load(Stream(format"PNG", io))`.

But there's also value in encouraging folks to define an internal (non-exported) `load_` function, for cases where we need to resolve ambiguities. Unless people prefer some alternative mechanism, of course.

src/error_handling.jl
```diff
@@ -0,0 +1,45 @@
...
+    message::UTF8String
+end
+Base.showerror(io::IO, e::NotInstalledError) = println(io, e.library, " is not installed.")
+
+function handle_error(e)
+	rethrow(last(first(e)))
+end
+
+function handle_error(e::LoaderError)
+
```
timholy: Unfinished?

timholy: Ah, I see that's your
> next step

rsrock:
Just so I'm clear on what I'd like to see here with the error handling, I'd like to be able to define a set of "acceptable errors", where I throw them explicitly when I'd like the next loader to take over.  The best example is something like `notindexed() || throw("QuartzImageIO", "Indexed images not implemented yet")`

In all other cases, I'd like the error to appear immediately, and in it's usual noisy way.  That way I get the bug reports.

Still not clear on why the `LoadError` (not `LoaderError`) in the Travis errors.

timholy:
> The best example is something like notindexed() || throw("QuartzImageIO", "Indexed images not implemented yet")

From a user-perspective, wouldn't it be better to fall back to something else (e.g., ImageMagick) when that happens? One could presumably set it up so that if it prompts you to install ImageMagick and you refuse, then it throws the original error. Would that make you happy?

src/loadsave.jl
SimonDanisch: Sadly not... It takes 12s on windows and allocates 8mb... Also it errors for mysterious reasons... I suppose it errors when it can't find it in METADATA;)

src/loadsave.jl
SimonDanisch: Totally agree :)

src/loadsave.jl
SimonDanisch: I think loading a library at runtime is already bad enough, `import` makes this problem a little less problematic, since it at least doesn't mess with your namespace.
Also it nicely removes the need for the `load_`.
The `Stream` example is problematic indeed. 
But I'm quite afraid of the mess that would follow by having two load functions.
How would we formulate clear guidelines for devs, when to use which!?
Since you can never be sure that there is another library popping up for the same format, I'm afraid the final recommandation will turn out to be `always use load_`.
So I'd rather force users to do `ImageMagick.load(Stream(...))`, or do some smart dispatch in FileIO. Shouldn't this work just fine:
```Julia
#in FileIO
function load(s::Stream{Formatted{S}}) 
library = checked_load(S)
library.load(s)
end

Comment (timholy) [inline src/loadsave.jl]: > How would we formulate clear guidelines for devs, when to use which!?

See #16 for one attempt to do exactly that.

If we _don't_ overload `load` and `save`, then `IJulia` will have to contain logic like
```
if isdefined(Main, :QuartzImageIO)
    @eval QuartzImageIO.save(Stream(format"PNG", $io), $data)
elseif isdefined(Main, :ImageMagick)
    @eval ImageMagick.save(Stream(format"PNG", $io), $data)
elseif ...
end
```
which is _not_ desirable.

There is at least one alternative to `load` and `load_`: define `save(format"ImageMagick:PNG", ...)`.

**Comment by SimonDanisch** [comment]
That's why I started pushing the errors into a list so that we can resolve errors globabaly after we know what doesn't work in what way.
In other words, I agree with Tim Holy and already did the first steps to support this.

@rsrock I think we should go another route for your problems with QuartzImageIO. For users, errors should just pop up when we can be sure, that nothing really works.
E.g. we could implement some switches to define different error levels, or turning IO libraries on and off.

File: src/loadsave.jl (hunk 3 above)
Comment (timholy) [inline src/loadsave.jl]: Your example won't work because `checked_import` takes the package name, not the format, as the input.

Comment (SimonDanisch) [inline src/loadsave.jl]: Well yeah, was a little quick there, but this is easy to fix right? And like that you wouldn't have to do the hacky stuff in IJulia.

Comment (timholy) [inline src/loadsave.jl]: I'm not sure I fully understand, so let me just lay out the factors I'm concerned about:
- QuartzImageIO is to be preferred for OSX users, but it fails on 20% of images. It needs to dispatch to ImageMagick in such cases.
- Users should have good control over how loads happen simply by selection of installed packages
- There shouldn't be too much of a performance hit for users who want to load lots of small files. I think our `query` mechanism should be pretty fast, but the `try/catch` logic (when it comes to modules) will have some overhead. I was hoping that by overloading `load(format"PNG", filename)` we'd directly dispatch to the correct function.

The last consideration is least important, because I haven't done any benchmarking and don't know if there's anything to worry about.

Comment (SimonDanisch) [inline src/loadsave.jl]: 1.) solved by iterating through loading libraries and chosing the one that works
2.) E.g. `ImageMagick.load` + some switches should solve this
3.) That's what FileIO should handle?! Or are you refering to cases where FileIO can't recognize the format and you wish to call ImageMagick, which has better recognition? In that case, since I'm in favor of not exporting load, ImageMagick could just define `load(path::AbstractString)`. You don't need to append a `_` but append `ImageMagick`, which I think is semantically more sound.
4.) Yeeah... Not sure about that one, but worst case scenerio is to recommend using `ImageMagick.load` where it matters.

Comment (rsrock) [inline src/loadsave.jl]: I agree with Tim's list here.  I would add that point 2 helps with point 1.  On the 20% of images that fail on QuartzImageIO, 15% of those are just not implemented (either temporarily, or permanently).  The remaining 5% are bugs that I want to hear about.

If the user hits one of those 5%, they should 1) file an issue, and 2) change `using Images, FileIO, QuartzImageIO, ImageMagick` to `using Images, FileIO, ImageMagick` so they can keep working while waiting for the fix.

This is pretty much what we had before, just much more elegant than asking the user to comment out a line in Images.

Comment (SimonDanisch) [inline src/loadsave.jl]: What solution would entail the commenting out?
I'd like to warn about bugs and then just continue loading the file if possible. I think that guarantees the best user experience. For debugging please see my other comment.
From what FileIO does right now, it should alway just be `using FileIO` in my opinion...

**Comment by rsrock** [comment]
Sorry to be a pain, this package is great work.  Hats off to both of you.  However, I strongly disagree that we should wallpaper over errors in the name of a good user experience. An error is an error, and needs to be dealt with right away.  An escape hatch is a fine thing to have as mentioned above https://github.com/JuliaIO/FileIO.jl/pull/43#discussion_r41090961& https://github.com/JuliaIO/FileIO.jl/pull/43#issuecomment-145253756& https://github.com/JuliaIO/FileIO.jl/pull/43#issuecomment-145280029
... to let the user continue work with a quick fix.  But there should still be an error, not a warning. 

Here's a clear issue that I can practically guarantee will pop up at some point.  Currently, IM and QIIO disagree about the types of certain RGB files.  Let's say I develop a library against QIIO, and then at some point, pull down Julia `0.5+3774` which breaks QIIO. FileIO will hand off to IM, which will change the type of these RGB files.  That's a scary scary Heisenbug.

This is not a theoretical issue.  Because I cannot disable ImageMagick in any way at the moment, the Travis tests for QuartzImageIO actually tests the combination of QuartzImageIO and ImageMagick.  If you extend FileIO to catch QIIO errors and turn them into warnings, Travis will gleefully report that the tests pass, even when they didn't.

File: src/loadsave.jl (hunk 3 above)
Comment (timholy) [inline src/loadsave.jl]: Yes, I'm in favor of it just being `using FileIO`, and the user not having to worry about which package gets used. (`using Images` is necessary only if the user also wants to call functions in `Images`.)

I guess part of my issue here is that, if we use @SimonDanisch's strategy, there is absolutely no reason to make the format a type-parameter: we're basically "dispatching" by dictionary-lookup and try/catch, in which case `DataFormat` should be defined
```jl
immutable DataFormat
    format::Symbol
end
```
so we reduce our compile burden.

My original intention was that dispatch would be direct, once you knew what format you wanted (and users could control that format with `format"FMT"`) and the relevant package was loaded. In particular, you could load or save a file without needing to, e.g., check what package folders are present in the user's path each time you want to load another file of the exact same format. But there are problems, chief among them that there can be only one exported `load(f::File{format"FMT"})` function, and the last one to be loaded always wins (and no module-scoping can change that). The "escape hatch" I envisioned was the unexported `load_` mechanism discussed in #16 (it really would have been better if that PR had gotten more serious discussion earlier on...), and even that's only partial because it doesn't solve the "oops, I just replaced `QuartzImageIO`'s PNG-loader with one from `ImageMagick`". Just about the only solution I see to that problem is to have ImageMagick be considerate and not stomp on previous versions if they are already defined. For example:
```
function __init__()
    if !isdefined(Main, :QuartzImageIO)   # or could use !applicable(load, File(format"PNG", "dummy")) as a general sentinel?
        eval(Expr(:import, :IMBindings))
    end
end
```
The internal `IMBindings` module would define the globally-exported versions,
```
FileIO.load(...) = ImageMagick.load_(...)
```
You can see a more complete example this kind of design in ProfileView (see its interaction with ProfileViewGtk through their respective `__init__` functions). If it seems a little complicated, blame package precompilation and the fact that you can't count on the user's system having the same set of packages loaded at usage time as s/he did at compile time.

So, I think we're faced with a choice:
- Allow direct dispatch of the type I originally designed: FileIO gets out of the way once the appropriate package is already loaded, except possibly to query the magic bytes when the user doesn't specify the format directly. The downside is that life gets considerably more complicated for developers in situations like the QuartzImageIO/ImageMagick format conflict, both in avoiding stomping on each others' methods and also in knowing what to do if one package has an error and would rather defer to another one.
- Switch to Simon's "dispatch"-by-dictionary-lookup-and-try/catch. While arguably less julian, it does have the major advantage of not needing any kind of cooperation among "competing" loaders/savers---it relies on FileIO to know "everything," so individual loaders/savers can just throw an error when they get confused.

Right now we have an awkward hybrid of these two strategies. Let's try to decide on a course of action soon, and get on with it. I can get on board with your strategy, @SimonDanisch, but if we do that we should start from the ground up (with the definition of `DataFormat`).

SimonDanisch: > FileIO gets out of the way once the appropriate package is already loaded

I'm not sure to what extend you mean this. You will always need to call `FileIO.load(::AbstractString)`, and do the query, right? If the library is already loaded, it actually doesn't get to the point to look it up in the path with this PR. It almost immediately dispatches it to the library in that case. 
There was never a direct dispatch in FileIO, was there? Even in the original design you needed to drop down to the actual IO library and do a `forced load` by knowing the correct `format`... Which you still can, right!? Okay, because of `import` vs `using`, you don't suddenly have them in your namespace. I sort of feel like this is a good thing though :D 
I'm not sure how the dictionary solution would look like without multiple dispatch...
What would you look up, when not a datatype to dispatch on? Or would you assume, that every IO library only has one function? This might be true for C-libraries but not for e.g. MeshIO which is pure Julia.

Right now it seems a little difficult to agree on anything... I'll sleep over this and see what the morning brings :)

SimonDanisch:
@rsrock no worries, we need to have this discussion =)
I would like to treat RGB{U8} vs RGB4 as a bug, so it shouldn't be something to base a design on.
I don't agree, that this will lead to scary heisenberg bugs... If one day QuartzImageIO decides to stop working you would get a warning (maybe even with a stack trace) which will coincide with an error with exactly the image you loaded...I'm not sure how this is any better than an error, looking up the treatment, restarting julia, changing a couple of lines to switch out QuartzImageIO with ImageMagick and then getting a different, non-IO error, even though everything seems to work just fine.

And for debugging and tests, I'd like to offer switches, as pointed out. For travis, you wouldn't even have ImageMagick installed, if it weren't for the REQUIRE in Images ( which we want to change, I think)

timholy:
Hmm, this is indeed a difficult issue. I think my original vision of FileIO was that once you picked a package to load, FileIO was done. If you picked QIIO and QIIO decided it couldn't handle the request, it was responsible for handing-off to to some other package (IIRC the main cause of failure was that the OSX library sometimes returned an empty vector for the colorspace?). That at least leaves QIIO in charge.

In the recent changes, any failure leads to it trying another package (if any is available). Thinking of it in light of your comments, I agree that this is probably too far.

src/loadsave.jl
timholy: If the library is loaded, then yes, `load(filename)` queries the magic bytes, but `load(File{format"FMT"}))` never goes through all the stuff in `loadsave.jl` (or didn't when I first wrote it). The distinction is that the user only has to specify the _format_, not the _package_. Whereas if we don't overload `load`, then the user has to specify _both_.

SimonDanisch: I'm not sure if we aren't deep into bikeshedding here.
From my benchmarks, there is no measurable difference between calling a function directly and calling it via a try catch, which dispatches immediately to the real function when the module is already loaded.
This means, we can always make your `load(File{format"FMT"}))` work within FileIO without qualifying the IO library or loosing performance.

timholy: It would likely show up for small formatted reads from an `IO`, say a dedicated format with a few bytes for which many can be packed into a single stream. But I agree it's reasonable that the performance shouldn't be a deciding factor in the design for "typical" cases, and it would be great to get past that.

> I'm not sure how the dictionary solution would look like without multiple dispatch...

I guess what bothers me is that we have this awkward hybrid right now. 95% of the time, all the important "dispatch" decisions are being made by dictionary-lookup and module-scoping. The only place dispatch is being used is _within_ a loader/saver module, for those (likely to be few?) cases where a module supports multiple formats. 

For those that do support multiple formats, we can either do the dispatch internally in those modules, or use
```
if sym == :PNG
    # do something
elseif sym == :TIF
    # do something else
end
```
The point being that if 95% of all of our dispatch is going to be via dictionary lookup and module scoping, there seems little point in creating hundreds of new types (all the new `DataFormat{sym}` types) and compiling a brand-new version of all of FileIO's functions for each one.

timholy:
Actually, come to think of it: we can do what @rsrock wants and what @SimonDanisch wants if instead of
```jl
    try
        Library = checked_import(library)
        return Library.load(q, args...; options...)
    catch e
        push!(failures, (e, q, args, options))
    end
```
we make it
```jl
    try
        Library = checked_import(library)
        return Library.load(q, args...; options...)
    catch e
        if isa(e, LoaderError)
            push!(failures, (e, q, args, options))
        else
            rethrow(e)
        end
    end
```
That way, QIIO can choose to either abort or ask for a fallback just by choosing whether to throw `LoaderError` or a different exception type.

src/loadsave.jl
timholy: Here's one practical example: if we _didn't_ have all the `DataFormat{sym}` types, then you probably wouldn't care whether you precompiled ImageMagick.jl or not. It's really the presence of [these lines](https://github.com/JuliaIO/ImageMagick.jl/blob/bf769a154c32c7d410f9508f70ae2f9fb9963a05/src/ImageMagick.jl#L60-L68) that make it take any time whatsoever; the amount of code in that repo is otherwise quite small.

SimonDanisch: I'm not getting what's so bad about the hybrid solution. It seems elegant to me as it gives us the best from two worlds :D 
If we don't export, we could also just do `load(::Formatted)` in `ImageMagick`, solving your second concerns.
I think we shouldn't worry about the compilation for the method, since loading different types requires different methods. With multiple dispatch, we have the advantage to call them all the same, while they're logically different functions. So having LLVM emit different functions for different formats is all well and good in my view.
In contrast, I consider it as bad design, that all formats are meshed up in one function like in `ImageMagick`, but it's just how it is when you have one monolithic c-library handling everything for you. So it's fine in that case and I think we have everything at our hands to handle this case well. I don't see why this one pathological case is an argument for dropping the whole dispatch design.

SimonDanisch:
We can do all sorts of things to make everyone happy...
We could also have a debug switch at that place. I don't think we really need to nail this right away, since it's really easy to just switch out a few lines of code to change the behaviour into one or the other direction.

src/loadsave.jl
timholy: I disagree that it's elegant to create hundreds of types that you won't use.  But moving forward is definitely a big priority, so if that's what you want to do in order to finish this PR, go for it.

timholy:
I think it's fair to say that we can't really move forward with tagging a new Images until this is resolved, but it can be done via a separate PR (one that I can write, if you don't want to).

SimonDanisch:
What I meant is, that all solutions are not totally terrible ;) So we should probably just implement one and see how people like it and act upon that feedback.

timholy:
Yeah, definitely getting something merged to master so it's easier to start playing with it would be great.

SimonDanisch:
Cool ;) 
I've the feeling that I showed, that all concerns can be handled with my approach.
Should I just go ahead and implement the rest? Or are there still staggering problems?

timholy:
If you're willing, by all means go for it.

rsrock:
Yes, go for it.  We can try a few variations on this theme once something is working on master.

Another approach that I thought about last night is to have loaders return a `Nullable{Image}`, which can then be checked and unwrapped if successful.  Just throwing that out there as food for thought.  I might take a stab at something like that at some point to see how it works.

SimonDanisch:
The error management know warns with stacktrace if there are still loading libraries left and if nothing works rethrows occured errors. It's easy to add more fine tuned handling, as demonstrated by handling `NotInstalledExceptions`, which will ask the user if it's wanted to install the library.

I was wondering what's missing for the final push!
- [ ] reading from Uint8 arrays
- [ ] getting the error management accepted 
- [ ] more tests for Images, so we can be sure there won't be regressions for the user

src/error_handling.jl
```diff
@@ -0,0 +1,73 @@
...
+function handle_error(e::NotInstalledError, q)
+    println("Library ", e.library, " is not installed but can load format: ", q)
+    println("should we install ", e.library, " for you? (y/n):")
+    input = chomp(strip(readline(STDIN)))
+    if input == "y"
+        info("Start installing ", e.library, "...")
+        Pkg.add(string(e.library))
+        return false # don't continue
+    end
```
timholy: Possibly add `else println("Not installing ", e.library) end`? I'm wondering about the case where the user accidentally misses the "y" key.

src/error_handling.jl
```diff
@@ -0,0 +1,73 @@
...
+end
+
+handle_error(e, q) = rethrow(e)
+
+function handle_error(e::NotInstalledError, q)
+    println("Library ", e.library, " is not installed but can load format: ", q)
+    println("should we install ", e.library, " for you? (y/n):")
+    input = chomp(strip(readline(STDIN)))
+    if input == "y"
```
timholy: `lowercase(input)`?

src/error_handling.jl
timholy: This is really nice, by the way.

src/error_handling.jl
```diff
@@ -0,0 +1,73 @@
...
+immutable NotInstalledError <: Exception
+    library::Symbol
+    message::UTF8String
+end
+Base.showerror(io::IO, e::NotInstalledError) = println(io, e.library, " is not installed.")
+
+@doc """
+Handles error as soon as they get thrown while doing IO
+""" ->
+function handle_current_error(e, islast)
```
timholy: So for automated travis testing, the right approach is for (e.g.) QuartzImageIO to redefine this function:
```jl
FileIO.handle_current_error(e, islast) = throw(e)
```
? CC @rsrock.

timholy:
With the ability to add https://github.com/JuliaIO/FileIO.jl/pull/43#discussion_r41369656 to a package's `runtests.jl` script, I'm quite happy with this.

Not sure what you mean by

> reading from Uint8 arrays

Are you referring to [this commit](758ec3b1dd63d04fb90c0d43bf6065be2739dee1)? (Already in ImageMagick.jl.)

src/error_handling.jl
```diff
@@ -0,0 +1,80 @@
...
+    bt = catch_backtrace()
+    bts = sprint(io->Base.show_backtrace(io, bt))
+    message = islast ? "" : "\nTrying next loading library! Please report this issue on Github"
+    warn(string(e, bts, message))
+end
+handle_current_error(e::NotInstalledError) = warn(string("lib ", e.library, " not installed, trying next library"))
```
timholy: This needs a 2nd argument, I believe.

SimonDanisch: Ah yeah, didn't notice it, since it's overshadowed by the other method ;)

SimonDanisch:
> Are you referring to this commit? (Already in ImageMagick.jl.)

The link doesn't work ;) But I quite likely oversaw this. But it doesn't work together with FileIO, right? When we make it work with FileIO, I'd rather like to call it decode instead of `load`...
`Load(::Vector{Uint8})` seems weird... Or is it just me!?

timholy:
Sorry, proper link is https://github.com/JuliaIO/ImageMagick.jl/commit/758ec3b1dd63d04fb90c0d43bf6065be2739dee1

I can definitely go with `decode`.

SimonDanisch:
@rsrock What about having a `kw_arg` to tell load/save to force using a certain library? That seems to be the sanest approach.

src/error_handling.jl
```diff
@@ -0,0 +1,80 @@
...
+        continue_ || break
+    end
+end
+
+handle_error(e, q) = rethrow(e)
+
+function handle_error(e::NotInstalledError, q)
+    println("Library ", e.library, " is not installed but can load format: ", q)
+    println("should we install ", e.library, " for you? (y/n):")
+    while true
```
rsrock: What if there's no user present to see this? Travis, or a run on a cluster?  I suppose we can override this function, but how about continuing after a timeout with no response?

src/error_handling.jl
rsrock: I like it.  This approach looks very flexible.

SimonDanisch:
The time out implementation is obviously not working :D 
I'm not sure how to implement this. How do I interrupt `readline()`?

timholy:
Maybe just wrap this in an `isinteractive()`?

SimonDanisch:
Ha too easy... That seems to be a good enough solution for now!

SimonDanisch:
- [x] reading from Uint8 arrays (although deeper `FileIO` integration would be nice)
 - [x] getting the error management accepted
 - [ ] more tests for Images, so we can be sure there won't be regressions for the user (this is pretty ill defined... I wouldn't really know where to start. Maybe tests are okay as they are)
 - [ ] remove `ImageMagick` && `QuartzImageIO` from Images `REQUIRE`
 - [ ] make it possible to force a certain library for `load`/`save`

timholy:
> more tests for Images, so we can be sure there won't be regressions for the user (this is pretty ill defined... I wouldn't really know where to start. Maybe tests are okay as they are)

I'm not sure either. Let's go with what we have.

SimonDanisch:
> make it possible to force a certain library for load/save
I will handle this in another PR, so this should be good to merge!

PR #45, opened 2015-10-17
Report the library that triggered the error

rsrock:
Before this change, it was unclear exactly where to file Github issues.  This expands the error message to include the library with the error along with the instructions.


timholy:
Thanks!

PR #48, opened 2015-12-10
Changes AVI detection to a function that ignores WAV files

### LINKED ISSUE #47: `RIFF` header used for `.avi` and `.wav` files
Both `.avi` and `.wav` files use `RIFF` as their magic bytes, followed by 4 bytes of size information and then `WAVE` or `AVI`.

What's the general strategy for nested file types like this? Should there be a `format"RIFF"` loader that then dispatches to separate `format"WAV"` and `format"AVI"` loaders? Another option would be to fall back to file type if the magic bytes conflict, but that doesn't help you e.g. if you have a generic container type that can have different contained file types.

ssfrr:
fixes #47


ssfrr:
hold off on this, there's a typo in there and I'm digging into why the tests didn't catch it

ssfrr:
I see, it looks like FileIO doesn't use the given detection function unless there are multiple file types with a matching extension, so it wasn't getting called. whoops. Should be fixed now.

timholy:
Thanks!

PR #50, opened 2015-12-21
load and save now supported on OS X

cormullion:
As suggested [here](https://github.com/JuliaIO/QuartzImageIO.jl/issues/15), a pull request. Unable to do any tests, since I don't know enough Julia to understand the test suite here. :)

(I followed the instructions [here](http://docs.julialang.org/en/release-0.4/manual/packages/#making-changes-to-an-existing-package). First time for everything. Seems to work!)

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -37,42 +37,42 @@ add_format(
     format"TGA",
     (),
     ".tga",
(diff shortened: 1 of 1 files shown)


SimonDanisch:
I guess we can merge this, even though there are still 2 issues with it!

PR #58, opened 2016-04-21
Add `FLAC.jl` into registry for `.wav` and `.flac`

staticfloat:


src/registry.jl
```diff
@@ -99,6 +99,20 @@ add_format(format"OFF", "OFF", ".off", [:MeshIO])
 
 
 
+### Audio formats
+function detectwav(io)
+    seekstart(io)
+    magic = ascii(read(io, UInt8, 4))
+    magic == "RIFF" || return false
```
SimonDanisch: This is probably better as:
```Julia
magic = read(io, UInt8, 4)
magic == b"RIFF" || return false
```
For speed and so that we don't get ascii conversion errors;)

staticfloat: Done in 66233e2

SimonDanisch:
great, thanks a lot :)

PR #59, opened 2016-04-22
Use `WAV.jl` for `.wav` files

staticfloat:
It turns out `FLAC.jl`'s `.wav` handling is worse than I thought, so I've switched this over to `WAV.jl`, what a surprise.

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -108,7 +108,7 @@ function detectwav(io)
 
     submagic == b"WAVE"
 end
-add_format(format"WAV", detectwav, "wav", [:FLAC])
+add_format(format"WAV", detectwav, ".wav", [:WAV])
 add_format(format"FLAC","fLaC",".flac",[:FLAC])
 
 


src/registry.jl
```diff
@@ -108,7 +108,7 @@ function detectwav(io)
 
     submagic == b"WAVE"
 end
-add_format(format"WAV", detectwav, "wav", [:FLAC])
+add_format(format"WAV", detectwav, ".wav", [:WAV])
```
SimonDanisch: Should this rather be:
```Julia
add_format(
    format"WAV",
    detectwav,
    [".wav", ".WAV"],
    [:WAV],
    [:FLAC]
)
```
Or is it really terrible? ;)

staticfloat: I have a PR to remove it from `FLAC.jl`.

SimonDanisch: Okay ;)

PR #61, opened 2016-05-10
Update to master syntax

andreasnoack:


yuyichao:
Ooops, I started #62 without realizing this one......

This is one of the cases where the type of the field matters since there are users (in this case, even other part of the package) that are indirectly (through field accessing function) depend on the concrete type of the field...

andreasnoack:
I've done the same thing way too often. I'll just close this one.

PR #64, opened 2016-05-15
some tweaks to Compat.String usage

StefanKarpinski:
The tests seemed to pass but produced a lot of scary looking noise.

### DIFF
--- a/docs/make_docs.jl
+++ b/docs/make_docs.jl
@@ -1,5 +1,5 @@
-using FileIO, Compat
-import Compat.String
+using FileIO
+using Compat; import Compat.String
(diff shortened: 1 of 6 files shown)


yuyichao:
This is breaking.

timholy:
Are you referring to the warning messages, like `WARNING: FileIO.NotInstalledError(:NotInstalled,"")` and the `Trying next loading library! Please report this issue on the Github page for LoadTest1`? Those are deliberately testing error conditions. Perhaps we need to have a way to capture them (if we don't already) so they don't frighten the user.

FYI, on julia master `Pkg` tests cause similar noise, so this isn't unprecedented.

PR #68, opened 2016-06-13
add abaqus inp register

KristofferC:
Ref https://github.com/JuliaIO/MeshIO.jl/pull/20

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -188,3 +188,5 @@ function detect_stlbinary(io)
 end
 add_format(format"STL_ASCII", detect_stlascii, [".stl", ".STL"], [:MeshIO])
 add_format(format"STL_BINARY", detect_stlbinary, [".stl", ".STL"], [:MeshIO])
+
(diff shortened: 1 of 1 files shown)


timholy:
I gather there's no reasonable way to detect such files by anything other than file extension?

KristofferC:
Perhaps you could look for a line that starts with `*Part` somewhere in the beginning.. But this is ASCII files so no magic bytes and different exporters to this format are not really consistent..

timholy:
Sounds like we should merge as-is, then.

KristofferC:
There is a bit of catch 22 here in which you can't really get tests to pass on the reader package side until the PR in `FileIO` is merged (and tagged), but one probably shouldn't merge the PR here until the reader one is merged. I guess one has to just accept some lag.

timholy:
You can test the reader package by putting `Pkg.checkout("FileIO")` into the `.travis.yml` script. Best to do this with a separate commit. Once you're satisfied it works, you can rebase the reader PR, dropping that change to `.travis.yml`, and merge despite the complaints from Travis. Then we can tag both packages.

PR #69, opened 2016-06-27
Fix del_format when magic bytes are generated by function and add rel…

JohnHBrock:
…ated tests


timholy:
Nice catch, thanks!

ssfrr:
Is there a release planned in the near future with this? @JohnHBrock submitted a PR to LibSndFile.jl that  is triggering this issue, so my CI won't be happy until this fix is released.

PR #75, opened 2016-08-24
Register R data format

alyst:
The PR registers `RData` package for loading R data format files (`.rda` or `.rdata` extension).

Reading R data files is currently being moved out of [DataFrames](https://github.com/JuliaStats/DataFrames.jl) into [RData](https://github.com/JuliaStats/RData.jl) (see JuliaStats/DataFrames.jl#1031) and `read_rda()` function is deprecated in favor of `FileIO` mechanism. It would be nice if, by the time RDA-stripped version of DataFrames is tagged, `FileIO` knows about ".rda" and can guide the user to use "RData".

Btw, are extensions case-sensitive? Should I also add ".Rdata" or it would be associated automatically?


nalimilan:
> Btw, are extensions case-sensitive? Should I also add ".Rdata" or it would be associated automatically?

No idea, but if so it should be `.RData`.

alyst:
It looks like extensions are case-sensitive, so I added ".RData". However, FileIO doesn't recognize ".RData" files (RStudio sessions) neither based on extension nor the magic-checking method (I wonder how it works, since `load(<filename>)` do not seem to call `detect_rdata()`).

SimonDanisch:
>Btw, are extensions case-sensitive? Should I also add ".Rdata" or it would be associated automatically?

We don't do anything in FileIO automatically in that regard. So far my rule of thumb was to just add all common variation explicitely. E.g. `["jpg", "JPG"]`! That way it's also more transparent for the reader of `registry.jl`.

SimonDanisch:
Seems to work fine for me after removing this line:
https://github.com/JuliaStats/RData.jl/blob/master/src/RData.jl#L97
Would be nice to add a test with a minimally sized file!

alyst:
@SimonDanisch Sure, I will remove `add_format()` from RData when/if this PR gets in.
How exactly do you want to add a test? Are there some examples I can use for reference?

SimonDanisch:
In the test folder are some examples:
https://github.com/JuliaIO/FileIO.jl/tree/master/test/files
And e.g.:
https://github.com/JuliaIO/FileIO.jl/blob/master/test/query.jl#L281

alyst:
Are the tests ok? Is it ready for merging?

SimonDanisch:
Looks good to me!
thanks a lot! :)

alyst:
@SimonDanisch Thanks! Do you plan to tag the new version soon? Can the dependent packages refer to FileIO 0.1.2 in `REQUIRE` for RData format support?

SimonDanisch:
See: https://github.com/JuliaLang/METADATA.jl/pull/6135

alyst:
@SimonDanisch Thanks!

Issue #77, opened 2016-09-13
add `loadstreaming` and `savestreaming` functions


ssfrr:
IIUC `load` and `save` are generally one-off operations, i.e. `load("somefile.wav")` loads the full wave file into memory.

Sometimes though, you have huge files that you might not want to load into memory all at once. In my `LibSndFile` package I also export `loadstream` and `savestream` functions that instead return special stream types that you can read/write chunks of audio from.

@jongwook's MP3 package also has the same functionality, but it's not clear who should own those functions. It seems like it would make sense to have them in `FileIO` as they use all the same query tooling. In fact, we implement `load` and `save` as [wrappers](https://github.com/JuliaAudio/LibSndFile.jl/blob/8abfe8a5f62669f9d72b05ff1f7f9b91ce6ecfa2/src/LibSndFile.jl#L203-L212) around the streaming versions.

Do these functions seem like they could fit into the FileIO architecture? What are the video packages currently doing for this sort of thing?

timholy:
Why doesn't the existing `Stream` interface cover this?

Demo:
```jl
julia> using FileIO

julia> fn = Pkg.dir("FileIO", "test", "files", "magic1.tiff")
"/home/tim/.julia/v0.5/FileIO/test/files/magic1.tiff"

julia> s = open(query(fn))
FileIO.Stream{FileIO.DataFormat{:TIFF},IOStream}(IOStream(<file /home/tim/.julia/v0.5/FileIO/test/files/magic1.tiff>),Nullable{String}("/home/tim/.julia/v0.5/FileIO/test/files/magic1.tiff"))

julia> position(s)
0

julia> skipmagic(s)
FileIO.Stream{FileIO.DataFormat{:TIFF},IOStream}(IOStream(<file /home/tim/.julia/v0.5/FileIO/test/files/magic1.tiff>),Nullable{String}("/home/tim/.julia/v0.5/FileIO/test/files/magic1.tiff"))

julia> position(s)
4

julia> read(s, 100);

julia> position(s)
104
```

timholy:
(If this was just a documentation issue, we should fix that.)

ssfrr:
this was addressed in #78.

PR #78, opened 2016-09-28
Implement loadstreaming/savestreaming API

ssfrr:
I figured it would be easier to discuss #77 if we had some code to look at. This PR is not close to mergeable, it's mostly to play with how this might work and get some feedback from you guys on whether you think this is a promising direction.

Here's the context (copied from the issue):

IIUC load and save are generally one-off operations, i.e. load("somefile.wav") loads the full wave file into memory.

Sometimes though, you have huge files that you might not want to load into memory all at once, or streams that have no end (like an internet radio stream). In my LibSndFile package I also export loadstream and savestream functions that instead return special stream types that you can read/write chunks of audio from.

@jongwook's MP3 package also has the same functionality, but it's not clear who should own those functions. It seems like it would make sense to have them in FileIO as they use all the same query tooling. In fact, we implement load and save as wrappers around the streaming versions.

Do these functions seem like they could fit into the FileIO architecture? What are the video packages currently doing for this sort of thing?

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,10 @@ file format.
 Sometimes you want to read or write files that are larger than your available
 memory, or might be an unknown or infinite length (e.g. reading an audio or
 video stream from a socket). In these cases it might not make sense to process
-the whole file at once, but instead process it a chunk at a time. For these situations FileIO provides the `loadstreaming` and `savestreaming` functions, which return an object that you can `read` or `write`, rather than the file data itself.
+the whole file at once, but instead process it a chunk at a time. For these
+situations FileIO provides the `loadstreaming` and `savestreaming` functions,
+which return an object that you can `read` or `write`, rather than the file data
+itself.
 
 This would look something like:
 
@@ -69,7 +72,8 @@ do loadstreaming("bigfile.wav") audio
 end
 ```
 
-Note that in these cases you may want to use `read!` with a pre-allocated buffer for maximum efficiency.
+Note that in these cases you may want to use `read!` with a pre-allocated buffer
+for maximum efficiency.
 
 ## Adding new formats
 
@@ -173,7 +177,13 @@ automatically even if the code inside the `do` scope throws an error.)
 Conversely, `load(::Stream)` and `save(::Stream)` should not close the
 input stream.
 
-`loadstreaming` and `savestreaming` use the same query mechanism, but return a decoded stream that users can `read` or `write`. You should also implement a `close` method on your reader or writer type. Just like with `load` and `save`, if the user provided a filename, your `close` method should be responsible for closing any streams you opened in order to read or write the file. If you are given a `Stream`, your `close` method should only do the clean up for your reader or writer type, not close the stream.
+`loadstreaming` and `savestreaming` use the same query mechanism, but return a
+decoded stream that users can `read` or `write`. You should also implement a
+`close` method on your reader or writer type. Just like with `load` and `save`,
+if the user provided a filename, your `close` method should be responsible for
+closing any streams you opened in order to read or write the file. If you are
+given a `Stream`, your `close` method should only do the clean up for your
+reader or writer type, not close the stream.
 
 ```julia
 struct WAVReader
--- a/README.md
+++ b/README.md
@@ -168,8 +168,9 @@ end
 Note that these are `load` and `save`, **not** `FileIO.load` and `FileIO.save`.
 Because a given format might have multiple packages that are capable of reading it,
 FileIO will dispatch to these using module-scoping, e.g., `SomePkg.load(args...)`.
-Consequently, **packages should define "private" `load` and `save` methods, and 
-not extend (import) FileIO's**. 
+Consequently, **packages should define "private" `load` and `save` methods (also
+`loadstreaming` and `savestreaming` if you implement them), and not extend
+(import) FileIO's**.
 
 `load(::File)` and `save(::File)` should close any streams
 they open.  (If you use the `do` syntax, this happens for you
@@ -191,17 +192,19 @@ struct WAVReader
     ownstream::Bool
 end
 
-function read(reader::WAVReader, frames::Int)
+function Base.read(reader::WAVReader, frames::Int)
     # read and decode audio samples from reader.io
 end
 
-function close(reader::WAVReader)
+function Base.close(reader::WAVReader)
     # do whatever cleanup the reader needs
     reader.ownstream && close(reader.io)
 end
+
+# FileIO has fallback functions that make these work using `do` syntax as well,
+# and will automatically call `close` on the returned object.
 loadstreaming(f::File{format"WAV"}) = WAVReader(open(f), true)
 loadstreaming(s::Stream{format"WAV"}) = WAVReader(s, false)
-# FileIO has fallback functions that make these work using `do` syntax as well.
 ```
 
 If you choose to implement `loadstreaming` and `savestreaming` in your package,
@@ -215,7 +218,7 @@ function save(q::Formatted{format"WAV"}, data, args...; kwargs...)
 end
 
 function load(q::Formatted{format"WAV"}, args...; kwargs...)
-    savestreaming(q, args...; kwargs...) do stream
+    loadstreaming(q, args...; kwargs...) do stream
         read(stream)
     end
 end
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ try
         # process the chunk
     end
 finally
-    close(stream)
+    close(audio)
 end
 ```
 
@@ -64,7 +64,7 @@ or use `do` syntax to auto-close the stream:
 
 ```jl
 using FileIO
-do loadstreaming("bigfile.wav") audio
+loadstreaming("bigfile.wav") do audio
     while !eof(audio)
         chunk = read(audio, 4096) # read 4096 frames
         # process the chunk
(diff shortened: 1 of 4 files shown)


SimonDanisch:
Thanks! I'd love to have streaming support in FileIO!
I think current video packages are not integrated yet :(

timholy:
See https://github.com/JuliaIO/FileIO.jl/issues/77#issuecomment-250913914. If this adds functionality on top of what's already there, great, but let's at least address what's missing.

ssfrr:
The difference is that these streams produce higher-level formatted data, not just raw file bytes. For instance:

```julia
s = loadstreaming("myfile.ogg")
x = read(s, 1000) # returns 1000 frames of audio in a SampleBuf
close(s)
```

In this case the underlying bytestream on disk is ogg-formatted, but when I read from this formatted stream I get samples of audio. If the ogg file is very large (or infinite), I can read or process it in chunks. It also provides a unified API between opening streams to hardware capture/playback and working with files, or infinite network streams.

You can see some other examples of this API in [LibSndFile.jl](https://github.com/JuliaAudio/LibSndFile.jl), though in that library it's currently called `loadstream`. I think `loadstreaming` may be a better name.

timholy:
I see. That's not bad, in a multi-image file we could also load a single frame at a time. :+1: 

Regarding names, to me it seems that the distinguishing feature here (to avoid the kind of confusion that nabbed me) is the "chunk." `loadchunk`?

ssfrr:
I think `loadchunk` implies we're loading a single chunk. IMO the difference isn't necessarily the chunkiness of the reading (you could read 1 sample at a time), it's that the thing you get from your `loadX` operation is itself a stream. I suppose I would be open to `loadchunked`, but that makes it seem like the stream you get has some inherent chunkiness rather than being a continuous stream you can read from in any size you want.

I think from a user's perspective it's a reasonable convention that `load("somefile.mov")` gets them the whole video, but `loadstreaming("somefile.mov")` gives them a video stream that they can `read` from (and get e.g. a `Vector{Image}` or some `Video` object or something). Another way to think about it is that the function name indicates what sort of object is returned (the output), and the argument type indicates what sort of underlying data is being used (the input).

For my purposes I generally consider streaming to be the more fundamental operation, in the sense that

```julia
x = load("mysong.wav")
```

is equivalent to

```julia
str = loadstreaming("mysong.wav")
x = read(str)
close(str)
```

or even better:

```julia
x = loadstreaming("mysong.wav") do str
    read(str)
end
```

We could in fact define `load` and `save` fallbacks that are defined as above, so packages can just implement `loadstreaming` and `savestreaming` and get `load` and `save` automatically.

Though I'm sure there are data types where you have to read the whole thing before you can start doing anything useful, so in those cases maybe the relationship would be reversed.

To better demo the benefit, this is a demo that works on the current LibSndFile, except the `loadstream` changed to `loadstreaming`.

```julia
loadstreaming("srcfile.wav") do src
    savestreaming("destfile.wav", nchannels(src), samplerate(src), Float32) do dest
        while write(dest, float(read(src, 2048)) / 2) == 2048 end
    end
end
```

And you get an output with half the amplitude of your input, while only keeping 2048 frames in memory at a time (of course in real code we'd want to pre-allocate the buffer to reduce allocation).

ssfrr:
Just rebased and added some tests, should be ready for review.

There's some refactoring that could be done to remove some duplication, but I figured I'd leave it as-is for now to make the diff easier to read. I can either do the refactoring as part of this PR or a follow-up.

ssfrr:
used a little metaprogramming to reduce the duplication introduced in this PR.

README.md
```diff
@@ -132,6 +170,50 @@ automatically even if the code inside the `do` scope throws an error.)
...
+
+```jl
+function save(q::Formatted{format"WAV"}, data, args...; kwargs...)
+    savestreaming(q, args...; kwargs...) do stream
+        write(stream, data)
+    end
+end
+
+function load(q::Formatted{format"WAV"}, args...; kwargs...)
+    savestreaming(q, args...; kwargs...) do stream
```
timholy: `loadstreaming`

README.md
```diff
@@ -132,6 +170,50 @@ automatically even if the code inside the `do` scope throws an error.)
...
+given a `Stream`, your `close` method should only do the clean up for your
+reader or writer type, not close the stream.
+
+```jl
+struct WAVReader
+    io::IO
+    ownstream::Bool
+end
+
+function read(reader::WAVReader, frames::Int)
```
timholy: Is this a private `read` or `Base.read`? (FileIO's docs were written, I think, before the the whole "do you extend or not?" debate was settled, but we should certainly fix this.)

README.md
```diff
@@ -132,6 +170,50 @@ automatically even if the code inside the `do` scope throws an error.)
...
+    # read and decode audio samples from reader.io
+end
+
+function close(reader::WAVReader)
```
timholy: Likewise, clarify whether this is `Base.close` or a private `close`. (It's fine to explain all in just one place.)

Also, I notice your test implements `eof`, worth mentioning?

README.md
```diff
@@ -132,6 +170,50 @@ automatically even if the code inside the `do` scope throws an error.)
...
+    # do whatever cleanup the reader needs
+    reader.ownstream && close(reader.io)
+end
+loadstreaming(f::File{format"WAV"}) = WAVReader(open(f), true)
```
timholy: Add blank line before this one? Or does the "fallback" comment below also apply to `close`?

src/loadsave.jl
```diff
@@ -49,77 +48,145 @@ the magic bytes are essential.
...
+be read in chunks and processed, rather than all at once. Reading from these
+higher-level streams should return a formatted object, like an image or chunk of
+video or audio.
+
+- `loadstreaming(filename)` loads the contents of a formatted file, trying to infer
+the format from `filename` and/or magic bytes in the file. It returns a streaming
+type that can be read from in chunks, rather than loading the whole contents all
+at once
+- `loadstreaming(strm)` loads the stream from an `IOStream` or similar object. In this case,
+the magic bytes are essential.
```
timholy: Not quite sure what "are essential" means. Perhaps say "the format is inferred from magic bytes in `strm`"?

src/loadsave.jl
```diff
@@ -49,77 +48,145 @@ the magic bytes are essential.
...
+- `load(File(format"PNG",filename))` specifies the format directly, and bypasses inference.
+- `load(f; options...)` passes keyword arguments on to the loader.
```
timholy: `load`->`loadstreaming`? Also, is that really `File` or did you mean `Stream` (as below with `savestreaming`)?

src/loadsave.jl
```diff
@@ -49,77 +48,145 @@ the magic bytes are essential.
...
+save
 
+"""
+Some packages may implement a streaming API, where the contents of the file can
+be written in chunks, rather than all at once. These higher-level streams should
+accept formatted objects, like an image or chunk of video or audio.
+
+- `savestreaming(filename, data...)` saves the contents of a formatted file,
+trying to infer the format from `filename`.
+- `savestreaming(Stream(format"PNG",io), data...)` specifies the format directly, and bypasses inference.
```
timholy: So this produces an output `Stream`, not a file, right? Is `savestreaming(File(...))` also supported? (bypassing inference)

README.md
ssfrr: whoops, should be `Base.[read|write|close]`. I'll fix. re: `eof`, I'd love it if there were a well-defined and documented stream interface we could link to here. I can try at least adding a list of functions to implement here and maybe eventually migrate it to [the julia docs](https://docs.julialang.org/en/stable/manual/interfaces/#)

src/loadsave.jl
ssfrr: hah, copied that from the `load` docstring. I think the idea is that because there's no file extension the magic bytes are the only way to get the file type. I can clarify.

src/loadsave.jl
ssfrr: `loadstreaming` and `savestreaming` (and `load` and `save` I believe) should support `File`s and `Stream`s. For example, if I call `loadstreaming` on a stream coming in from a network, I should be able to `read` formatted data (e.g. audio frames or video frames) from the returned object, which will in turn read from the stream. calling `loadstreaming` on a `File` just opens that file and then handles it as a stream.

similarly `savestreaming` should handle both `File`s and `Streams`.

The new commit I just pushed adds more methods to the docstring. I also added them for `load` and `save`, which I believe should also support both.

src/loadsave.jl
ssfrr: clarified in new commit

README.md
ssfrr: re-ordered the comment to clarify this

README.md
ssfrr: I went down a bit of a rabbit hole trying to figure out a more formalized stream interface. My notes are [here](https://gist.github.com/ssfrr/73a232a5bc52f89d7bf2173f6917274d), but basically it's not super well-defined at this point. Once that's nailed down we can link to it in these docs.

ssfrr:
I just rebased and made edits based on your comments. Apologies this sat dormant for a while, I know that makes it tougher to review. Compared to what you looked at last time this is just documentation updates.

Tests seem to be passing on 0.6 but failing on nightlies due to some Nullable issue - probably unrelated to this PR?

ssfrr:
bump. 😃

SimonDanisch:
Well, if @timholy loves it I don't see why we shouldn't merge this ;) If no one is against it, I merge it tonight!

Issue #53, opened 2016-01-15
Infinite recursion on save("test.pgm")


mewo2:
I'm not sure if this is a bug in FileIO, or in the underlying Netpbm (or even a bug at all - I only noticed it because of a misplaced parenthesis). Anyway, one can induce an infinite recursion with the following code:
```julia
using Images
save("test.pgm")
```
It seems that in the absence of any data to save, the call at `loadsave.jl` line 92 is recursing infinitely.

SimonDanisch:
It's a bug! ;) It should be fixed at some point!

staticfloat:
What is the proper way to avoid this?  I'm trying to enforce the usage of certain parameters when `save()`'ing out a `.flac` file, but if I enforce a certain method signature, then we get recursive compilation which kills Julia.  :(

The best I can come up with is defining all my `save()` methods as `save(f::File{format"FLAC"}, data..., kwargs...)` so as to catch _all_ incoming method signatures, and then manually inspecting `data` and `kwargs`, which is kind of a code smell, in my opinion.  I think we should have some way of enforcing which module the `save()` function is being pulled from in `loadsave.jl` on line 92, so that we can actually fail to compile without hitting infinite recursion.

SimonDanisch:
Just to make sure that I understand correctly:
You have a certain signature with different kw_args/data, which is only supported by a certain Module, so you'd like to enforce using that Module?

staticfloat:
My example is that I defined `save(f::File{format"FLAC"},  data, samplerate)`, in an effort to force the user to provide a samplerate along with their filename and data in order to save a `.flac` file.  However, if the user calls `save("test.flac", [1,2,3])` the compiler wigs out because [this line](https://github.com/JuliaIO/FileIO.jl/blob/e10242de106c8c27823fb920fad787beef5aff26/src/loadsave.jl#L95) recurses infinitely.  This is because `FLAC` imports `FileIO` to overload `save()`, and so that line just ends up calling itself.

SimonDanisch:
Okay, got it!
Not really solving the problem, but would it make sense to make the sample rate part of the data?

SimonDanisch:
Okay so you shouldn't import FileIO.save/load into FLAC.jl...
And then in line 95, we should add:
```Julia
Library.save == FileIO.save && error("Library does not have a save method")
```
Can you confirm, that this solves it?

staticfloat:
[...]

PR #79, opened 2016-10-01
Fix stack overflow when load/save are missing

timholy:
On master, the following leads to a stack overflow:
```jl
julia> using FileIO

julia> module BrokenIO
       using FileIO
       end
BrokenIO

julia> add_format(format"BROKEN", (), ".brok", [:BrokenIO])
FileIO.DataFormat{:BROKEN}

julia> load(Stream(format"BROKEN", STDIN))
```
This arises because `FileIO` exports `load`, and since `BrokenIO` doesn't define its own internal `load`, `BrokenIO.load` just expands to `FileIO.load`. So `FileIO.load` ends up calling itself recursively. Moreover, because of the (unavoidable) type instability in `FileIO.load`, it's a very slow stack overflow to resolve, so it effectively results in a hang.

Now, one solution would be to recommend that packages do this:
```jl
import FileIO
using FileIO: File, Stream, @format_str
```
instead of `using FileIO`. But I believe it's more robust to take the route I've taken in this PR (for one thing, it will catch any preexisting packages without modification).

### DIFF
--- a/src/error_handling.jl
+++ b/src/error_handling.jl
@@ -8,7 +8,7 @@ immutable LoaderError <: Exception
     msg::Compat.UTF8String
 end
 Base.showerror(io::IO, e::LoaderError) = println(io, e.library, " load error: ",
-                                                 msg, "\n  Will try next loader.")
+                                                 e.msg, "\n  Will try next loader.")
 
 """
 `WriterError` should be thrown when writer library code fails, and other libraries should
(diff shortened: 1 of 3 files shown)


timholy:
Looking through old issues: this fixes #53,

SimonDanisch:
Great! lgtm

PR #80, opened 2016-10-01
Store the absolute path when querying or opening a Stream

timholy:
If this seems desirable, perhaps don't hit the merge button just yet. This is technically breaking, so we might want to first merge other queued PRs, tag a version, merge this, and then tag a minor version bump.

### DIFF
--- a/src/query.jl
+++ b/src/query.jl
@@ -279,7 +279,10 @@ function file!{F}(strm::Stream{F})
 end
 
 # Implement standard I/O operations for File and Stream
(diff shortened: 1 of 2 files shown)


timholy:
A new tagged release is queued, so we can merge this.

timholy:
Interesting observation: codecov is smarter than coveralls. codecov recognizes that a PR which tests all changed lines as passing, even if it results in a decrease in the coverage fraction.

PR #84, opened 2016-10-04
An alternative solution to the "missing method" problem

timholy:
See https://github.com/JuliaLang/METADATA.jl/pull/6619#issuecomment-251478364

This one relies on checking the module in which each method was defined. I worry this could be slow if the registry gets big and lots of modules (inadvisedly) extend load/save. On the other hand I can understand why legacy packages (EDIT: meaning, ones that existed before FileIO was around) are reluctant to stop exporting `load/save` and force their users to say `using FileIO`.

I am not sure how I feel about this, but on balance I suspect we should go this route.

CC @alyst

### DIFF
--- a/src/loadsave.jl
+++ b/src/loadsave.jl
@@ -79,7 +79,7 @@ function load{F}(q::Formatted{F}, args...; options...)
     for library in libraries
         try
             Library = checked_import(library)
-            if !isdefined(Library, :load) || Library.load == FileIO.load
+            if !has_method_from(methods(Library.load), Library)
                 throw(LoaderError(string(library), "load not defined"))
             end
             return Library.load(q, args...; options...)
@@ -96,7 +96,7 @@ function save{F}(q::Formatted{F}, data...; options...)
     for library in libraries
         try
             Library = checked_import(library)
-            if !isdefined(Library, :save) || Library.save == FileIO.save
+            if !has_method_from(methods(Library.save), Library)
                 throw(WriterError(string(library), "save not defined"))
             end
             return Library.save(q, data...; options...)
@@ -105,4 +105,19 @@ function save{F}(q::Formatted{F}, data...; options...)
         end
     end
     handle_exceptions(failures, "saving \"$(filename(q))\"")
+end
+
+function has_method_from(mt, Library)
+    for m in mt
+        if getmodule(m) == Library
+            return true
+        end
+    end
+    false
+end
+
+if VERSION < v"0.5.0-dev+3543"
+    getmodule(m) = m.func.code.module
+else
+    getmodule(m) = m.module
 end


alyst:
Would it make sense to have different names for user-facing `load/save` FileIO methods and extension API `load/save` methods (e.g. `loadformat()/saveformat()`)? Then the should be no danger of recursion.

timholy:
Note to self for when I have time: 0.4 needs `m.func.code.module`. A good reminder of how much Julia's internals are getting cleaned up with each release...

tkelman:
Do you expect to get this released today? If not, I'm going to upper-bound FileIO on all the packages this directly or indirectly broke (http://pkg.julialang.org/pulse.html - at least FreqTables, Gadfly, ImagineFormat, RDataSets, and VegaLite were due to this). Adding a direct dependency on `FileIO` and a `using FileIO` may be an easier fix than whatever's happening here, so new tags of the respective packages with that fix would not need upper bounds.

timholy:
I think it merits more discussion; I'd especially like @SimonDanisch's input, since he has thought about this a lot. So you should probably go with the upper bound.

Interestingly, this was supposed to be the *non*-breaking tag in preparation for #80 :cry: .

SimonDanisch:
Oh jeez, now I remember why I didn't fix the recursion problem right away...
What I like about forcing the user to explicitly use FileIO is, that it would establish that FileIO is the official way to load any kind of format, and that its the goto library if one doesn't know what package to use for format x.
What I don't like is the forcing part :P It should be possible to create meta packages where you only load a single library and everything works.
Should we go with this fix for now? If speed becomes an issue, we can still try to figure things out.

alyst:
I understand the motivation _"I don't know how to read this file, FileIO, please take care of that for me"_. OTOH the only purpose of `RData.jl` is to load R data files, so it would be a little bit confusing if just `using RData` is not enough to read the files of the specific format.

SimonDanisch:
I agree!

timholy:
Whether we should extend `load` or module-dispatch `load` is definitely one of the hardest decisions and something that has bothered me for a while. My original intention was to extend `load`, but Simon brought up some valid concerns about that approach. We never really have come to grips with all the corner cases, though.

@SimonDanisch, I'll try to fix the failure on 0.4. If you like this overall, then hit merge. If not, we need to come up with something.

timholy:
OK, since I hastily merged #80 we might need to do something complicated if we think the abspath change is worth a minor version bump. I'll push a new branch that omits that PR, tag it, and then tag a new minor bump against master.

tkelman:
What's the issue with abspath from #80? Are you expecting that to break things with 0.2.0?

tlnagy:
This fixed errors in Gadfly's tests. 👍

timholy:
I don't expect it to break anything, but in case anyone was counting on it
storing a relative path, I thought we should tag a new version.

PR #91, opened 2016-12-28
don't access String .data field

JeffBezanson:
Fixes package for jb/fasterstring branch. Should work for julia >= 0.4.


JeffBezanson:
Anybody know why AV seems not to be running? Anyway can this be merged?

SimonDanisch:
Thanks Jeff! :) I cant merge this from my phone, because tests are failing
( kinda amazing). Will merge when back home. I had the same issue with AV
recently, but I'd say its not that important to test this change on windows
for now!

PR #94, opened 2017-01-11
Fix depwarn and error on 0.6

yuyichao:


SimonDanisch:
Thanks! I have this resistance to `eval`, but I guess that's just how it works now...

PR #95, opened 2017-01-13
Update make_docs.jl deprecated takebuf_string

musm:


SimonDanisch:
thanks a lot!

REQUIRE
```diff
@@ -1,2 +1,2 @@
 julia 0.4
-Compat 0.7.19
+Compat 0.9.5
```
tkelman: users don't need to build the docs, this isn't necessary

tkelman: (I was wrong on this, was missed from #94)

PR #99, opened 2017-01-25
Register GSLIB file format

juliohm:
I am proposing (and coding) an extension to the file format introduced in http://gslib.com

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -110,7 +110,8 @@ add_format(format"PLY_BINARY", "ply\nformat binary_little_endian 1.0", ".ply", [
 add_format(format"2DM", "MESH2D", ".2dm", [:MeshIO])
 add_format(format"OFF", "OFF", ".off", [:MeshIO])
 
-
+# GSLIB/SGeMS format (http://gslib.com)
+add_format(format"GSLIB", (), [".gslib",".sgems"])
 
 ### Audio formats
 function detectwav(io)


juliohm:
Do I need to have it working before you can merge this?

SimonDanisch:
Yes, I thought that was the plan from your side as well ;)
There's not much use for a file format without a loader.

juliohm:
Hi @SimonDanisch , I finished the first version of the package, I just want to make sure everything is in place before I register it. The tests in Travis are failing, can you please review the code to see if I did something incorrectly?

https://github.com/juliohm/GslibIO.jl

Also, can you confirm that a user doesn't need to install my package to read/write the format, everything is managed indirectly by `FileIO.jl`? If that is the case I will remove the installation section from the README.

juliohm:
Can anyone give a hand on this?

SimonDanisch:
it says undefined variable error, which is because you're using an undefined variable in one of your last statements. It's a typo I assume.

juliohm:
@SimonDanisch I fixed the issue, I think now the only problem is that the format is not registered yet with FileIO, can you please confirm?

SimonDanisch:
Did you confirm it locally by running the tests on this branch?

juliohm:
I'm gonna double check locally...

-Júlio

SimonDanisch:
It'd be nice if you test your code yourself before committing :P 
But you're using it wrong. You must test it like this:
```Julia
using FileIO

array = load("test.gslib")
array = save("test.gslib")
```
Otherwise fileio can't do it's magic...

juliohm:
Fixed, tests pass locally now, can you please review?

SimonDanisch:
Okay, that error message looks more like it!
After merging, you can temporarily add a Pkg.checkout("FileIO") into your tests to be sure that things work before tagging a new FileIO version.

juliohm:
Thanks @SimonDanisch , I will checkout FileIO locally and let you know.

PR #104, opened 2017-02-02
Drop Julia 0.4 support

tkelman:
and stop extending `readbytes` which no longer exists on 0.6


musm:
looks like this needs a rebase

SimonDanisch:
Thanks :)

PR #106, opened 2017-02-04
Update FileIO.jl

musm:


src/FileIO.jl
```diff
@@ -1,16 +1,15 @@
-VERSION >= v"0.4.0-dev+6641" && __precompile__()
+__precompile__()
 
 module FileIO
+
 using Compat
-import Compat.String
+using Compat.String
```
musm: since we don't extend the `String` method

src/FileIO.jl
```diff
@@ -1,16 +1,15 @@
-VERSION >= v"0.4.0-dev+6641" && __precompile__()
```
musm: no  longer needed

PR #107, opened 2017-02-06
Fix error on 0.6

Keno:


SimonDanisch:
@timholy fine to tag, or do you think this will break some package?

tkelman:
maybe check the open pr's first?

timholy:
Seems fine to me.

Issue #90, opened 2016-12-24
The magic number for PCX files is too strict


magister-ludi:
The magic number for PCX format is set as `UInt8[0x0a,0x05,0x01,0x01]` in [registry.jl](https://github.com/JuliaIO/FileIO.jl/blob/master/src/registry.jl). The fourth element represents the number of bits per pixel, so that `load(filename)` can fail for images that will correctly load with `load(File(format"PCX", filename))`.
The [attached](https://github.com/JuliaIO/FileIO.jl/files/671615/testpcx.zip) 256-colour PCX file (zipped, so that it can be attached), with `0x08` as the fourth byte, is a simple example.

timholy:
Looks like really only the first byte is technically the magic byte; the others encode information that might be useful to a parser.

magister-ludi:
Strictly speaking, that's true. The third byte (encoding) should always be `0x01`. The second byte (version) is almost always `0x05`. In my own software, I have used the first three bytes for many years without a problem. However I don't know whether older PCX files (versions `0x00`, `0x02`, `0x03`, `0x04`) are still out there. or how much software actually supports earlier versions.
I think your options are
1. Use only the first byte, though I can understand any reluctance to do that.
1. Use the first three bytes, noting that images before version 3.0 (`byte3==0x05`) don't support magic number recognition
1. Remove the magic number for PCX format. In a practical sense, this is the situation that exists now, as the magic number is unreliable. Removal would make that explicit.
While option 2 would be the convenient for me, there are disadvantages with each of the options. It's really your call on which to choose, although I think the zeroth option (doing nothing) is not the best one.

timholy:
We also have a function version that is more flexible...see the code for HDF5.

timholy:
Though I'd be perfectly happy with option 2

magister-ludi:
Thanks for the tip about HDF5. For now it might be more overhead than I want, but I will keep it in mind. In the meantime, I've built a workaround into my own code.

I'm happy to close the issue now, and leave to you whether to make relevant changes to the FileIO code.

timholy:
If you're willing to tackle the fix, that would be a huge help. If not, let's at least leave this open until some motivated individual steps forward.

magister-ludi:
Happy to do that. I'll report when it's ready.

timholy:
[...]

PR #114, opened 2017-04-03
Modified magic number testing for PCX image files.

magister-ludi:
The magic numbers now match the magic numbers used by
ImageMagick. Some valid PCX files are still not identified
correctly. This addresses https://github.com/JuliaIO/FileIO.jl/issues/90.


timholy:
Thanks, @magister-ludi!

PR #119, opened 2017-05-15
Adding GZIP format and tests

hpoit:
Forgive me if I've missed anything.


SimonDanisch:
GZIP is a bit problematic, since a lot of other file formats use it under the hood (so they might actually have the same magic byte). But I think there is no harm in starting like this (as long as we have `.gz` as the file extension)!

SimonDanisch:
But you need to do the real integration in Libz to make this work!

hpoit:
SImon, I will! Thank you!

PR #126, opened 2017-06-09
Fix some save calls in the tests

davidanthoff:
This fixes some bugs that were introduced in #123.

In particular, there were a number of tests that used a ``save`` method with only a filename, and that collides with the stuff in #123. I think at the end of the day those test cases were artificial and can easily be changed like in this PR.

### DIFF
--- a/test/error_handling.jl
+++ b/test/error_handling.jl
@@ -7,10 +7,10 @@ println("these tests will print warnings: ")
     stderr_copy = STDERR
     rs, wr = redirect_stdin()
     rserr, wrerr = redirect_stderr()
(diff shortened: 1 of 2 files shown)


davidanthoff:
@simonbyrne That one check failure is some odd thing from codecov, I think this can be merged, it solves all the errors that I introduced in #123.

SimonDanisch:
I guess you wanted to ping me, @simondanisch? ;)
Excellent! Yes, these are very much artificial!

davidanthoff:
Yes, sorry!

PR #127, opened 2017-06-09
Add .gitattributes to fix line ending of rda test file

davidanthoff:
This fixes another test bug on Windows: if git is configured with the auto line ending setting, it will checkout the rda test file with CRLF line endings, and that breaks one of the tests. This setting makes sure that this file always has LR line endings. This bug didn't show up on appveyor because it uses a different git setting.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+test/files/minimal_ascii.rda eol=lf


davidanthoff:
I need to rebase this once #126 is merged and then the tests should pass.

SimonDanisch:
But wouldn't it be possible to actually get a file with CRLF line endings in the real world, so we should actually test against it and not fail that test?

davidanthoff:
The bug occurs in this [line](https://github.com/JuliaIO/FileIO.jl/blob/master/test/query.jl#L319), the position on Windows there would be 6 from the extra byte for the line ending. So I don't think there is a problem in the general package code, it is just that this specific test doesn't handle different line endings.

PR #130, opened 2017-06-11
Add some magic bytes

davidanthoff:
I'm still trying to find magic bytes for ``*.dta`` files.

I don't thing we should add magic bytes for the Excel files: both files use more generic file formats (OLE compound storage and zip files), so other files might get classified incorrectly as Excel files if we put magic bytes in.

This also removes the ``*.por`` file type. I couldn' get ReadStat.jl to read that kind of file, not sure what is going on there. Probably better to remove for now.

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -17,11 +17,14 @@ end
 add_format(format"RData", detect_rdata, [".rda", ".RData", ".rdata"], [:RData, LOAD])
 
 add_format(format"CSV", (), [".csv"], [:CSVFiles])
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Hm, right now we don't use the file extension to resolve magic byte ambiguities :(
So yeah, lets leave it out for now. I'm relatively sure that we should be fine like that for 99% of the cases, and if people start complaining, we can start doing the extra work of making our matching more robust!

SimonDanisch:
Thanks!

PR #138, opened 2017-07-03
Sd/fix137

SimonDanisch:
applicable savers shouldn't fall back to FileIO, since it will make save stack overflow for unkown formats.


tkelman:
should test this kind of thing

PR #139, opened 2017-07-07
Add JLD2 to data formats

simonster:
For now, I am registering the extension `.jld2` to avoid ambiguity with JLD when saving, although someday, after JLD2 receives some testing, we may want to reassign the `.jld` extension to JLD2.


simonster:
Thanks. And thanks for tagging!

Issue #144, opened 2017-07-28
Prioritization among loaders?


tlnagy:
I'm writing a loader for uMicromanager OME-TIFF files. These have the standard TIFF magic bytes and TIFF ending and are picked up by the standard TIFF loader. I tried inserting my custom detection function either above or below the TIFF `add_format` call and I couldn't get it to work unless I commented out that line. 

Is there any way to assign priority for loaders?

tlnagy:
@timholy any thoughts?

timholy:
I haven't worked on this package in ages, so I don't remember. You could check the implementation of `add_format` and see how it works. If it's a Dict, then the current answer is "no", but in that case we should change the implementation somehow (either to an ordered Dict, or assign a priority, or something).

tlnagy:
So my OMETIFF loader is up and running over at https://github.com/tlnagy/OMETIFF.jl, but currently has its own `loadtiff` function instead of depending on the `FileIO` infrastructure.

It looks like this might be a little complicated. My reading of `add_format` here https://github.com/JuliaIO/FileIO.jl/blob/master/src/query.jl#L78-L89 and query code here https://github.com/JuliaIO/FileIO.jl/blob/master/src/query.jl#L402-L426 is that

1) Magic bytes are sorted in front of magic functions
2) Magic bytes are *unique* to a specific format

This means that in my case, where I have identical magic bytes to already registered format, my loader would have lower priority than the default TIFF one. 

I think main way of solving this is to get rid of the uniqueness requirement and first sort on magic bytes and then on file name (`*tif` vs `*.ome.tif`). Since OME-TIFF is a simple extension of TIFF that's backwards compatible, there's no good way of distinguishing it quickly from normal TIFFs.

timholy:
OK, now I have a couple of seconds to rub together and can actually look at code.

[...]

PR #147, opened 2017-08-03
add support for loading OME-TIFFs

tlnagy:
See more discussion in https://github.com/JuliaIO/FileIO.jl/issues/144


tlnagy:
This needs to wait till OMETIFF.jl is registered in Metadata.jl and new tag is pushed.

tlnagy:
ffc.pcx is released under a CC0 license from here: http://www.fileformatcommons.com/pcx-file-format/

src/registry.jl
```diff
@@ -154,6 +147,21 @@ add_format(format"FLAC","fLaC",".flac",[:FLAC])
 
 ### Complex cases
 
+# Handle OME-TIFFs, which are identical to normal TIFFs with the primary difference being the filename and embedded XML metadata
+tiff_magic = (UInt8[0x4d,0x4d,0x00,0x2a], UInt8[0x4d,0x4d,0x00,0x2b], UInt8[0x49,0x49,0x2a,0x00],UInt8[0x49,0x49,0x2b,0x00])
```
timholy: Add `const` in front of this? Doing so should make `detecttiff` faster.

test/query.jl
```diff
@@ -286,16 +286,20 @@ end
...
-    q = query(joinpath(file_dir, "magic1.tiff"))
-    @test typeof(q) == File{format"TIFF"}
-    q = query(joinpath(file_dir, "magic2.tiff"))
-    @test typeof(q) == File{format"TIFF"}
+    q = query(joinpath(file_dir, "ffc.pcx"))
+    @test typeof(q) == File{format"PCX"}
     open(q) do io
         @test position(io) == 0
         skipmagic(io)
-        @test position(io) == 4
```
timholy: Now that each function has its own type, we could change how `skipmagic` works:
```julia
skipmagic(io, ::typeof(detect_ometiff)) = seek(io, 4)
skipmagic(io, ::typeof(detect_noometiff)) = seek(io, 4)
```
and then I think the original test would pass?

src/registry.jl
tlnagy: Sure thing!

test/query.jl
tlnagy: Where would it be best to put these `skipmagic` definitions? In `registry.jl` or `query.jl` where skipmagic is defined.

timholy: These specializations probably have to go in `registry.jl`, since we need `detect_ometiff` to be defined already.

tlnagy:
Hey @timholy, I made the changes you suggested. I think the failures on nightly aren't related to this PR.

timholy:
This is great, thanks!

PR #149, opened 2017-08-22
Register bedGraph format

CiaranOMara:
Please register [BedgraphFiles](https://github.com/CiaranOMara/BedgraphFiles.jl) for use with [.bedgraph](http://www.genome.ucsc.edu/goldenPath/help/bedgraph.html) files.


timholy:
Thanks!

PR #152, opened 2017-09-18
Register tsv and wsv files

davidanthoff:


src/registry.jl
```diff
@@ -19,6 +19,8 @@ end
 add_format(format"RData", detect_rdata, [".rda", ".RData", ".rdata"], [:RData, LOAD])
 
 add_format(format"CSV", (), [".csv"], [:CSVFiles])
+add_format(format"TSV", (), [".tsv"], [:CSVFiles])
+add_format(format"WSV", (), [".wsv"], [:CSVFiles])
```
timholy: The search hit I get for wsv seems to be a [graphical format](http://www.file-extension.org/extensions/wsv), is this really a standard? (tsv seems to universally come up with tab-separated value, so that seems more straightforward.)

davidanthoff: @mkborregaard any opinion?

I don't have any strong opinion about this, @mkborregaard brought this up over at https://github.com/davidanthoff/CSVFiles.jl/issues/10.

mkborregaard: The reason for suggesting `wsv` is that this is what DataFrames does on the stable version: https://juliadata.github.io/DataFrames.jl/stable/man/io/#Importing-and-Exporting-(I/O)-1
I think they're passing everything to CSV now though.
A more traditional extension for space-delimited could be `.prn`?

timholy: That also means [something else](http://www.frogmorecs.com/arts/what-is-a-prn-file/). This is the problem with file formats that don't have magic bytes: 3 letters is just not enough for disambiguation.

mkborregaard: It's fine for me not to have the space-separated format at all - I use mainly the tab-delimited. I just suggested this one for consistency with DataFrames. What's important for me is to have an alternative standard format than "csv", because "csv" means two different things by default in the former British colonies and the rest of the world (making it, in my opinion, possibly the worst choice for a standard format but I am aware that this convention predates julia).

timholy: Certainly we have to support csv (as bad as it indeed is), and I have no objections to tsv. If you remove the wsv, I have no objections to merging.

davidanthoff:
Alright, wsv stuff removed.

I've tagged a CSVFiles release that supports this: https://github.com/JuliaLang/METADATA.jl/pull/11273

So ideally we can merge things here and also tag a release?

Issue #141, opened 2017-07-22
Segfault due to stack overflow when calling `load` with arguments not supported by package


simonster:
If a package defines a `load` method for `File{format"X"}` but not for `Stream{format"X",<:IO}` (or if it defines `load` for one type of IO but not all types of IOs), then [this line](https://github.com/JuliaIO/FileIO.jl/blob/c43eacd733a8e6db4450c340f8f7d39fc6a7f5b2/src/loadsave.jl#L98) calls itself. See https://github.com/simonster/JLD2.jl/issues/19

timholy:
Do you see any non-breaking solution other than to collect a backtrace upon entry and determine the caller? (Or use introspection, but I think that's worse.)

timholy:
Fixed by https://github.com/simonster/JLD2.jl/pull/20

simonster:
Couldn't we just check whether `Library.load == FileIO.load`?

PR #154, opened 2017-09-20
Guard against StackoverflowErrors: prevent packages from extending FileIO.load or FileIO.save

timholy:
Follows up on https://github.com/JuliaIO/FileIO.jl/issues/141#issuecomment-317205080

### DIFF
--- a/src/loadsave.jl
+++ b/src/loadsave.jl
@@ -95,6 +95,9 @@ function load{F}(q::Formatted{F}, args...; options...)
             if !has_method_from(methods(Library.load), Library)
                 throw(LoaderError(string(library), "load not defined"))
             end
+            if Library.load == FileIO.load
+                throw(LoaderError(string(library), "module should not extend FileIO.load"))
+            end
             return eval(Main, :($(Library.load)($q, $args...; $options...)))
         catch e
             push!(failures, (e, q))
@@ -111,6 +114,9 @@ function save{F}(q::Formatted{F}, data...; options...)
             Library = checked_import(library)
             if !has_method_from(methods(Library.save), Library)
                 throw(WriterError(string(library), "save not defined"))
+            end
+            if Library.save == FileIO.save
+                throw(WriterError(string(library), "module should not extend FileIO.save"))
             end
             return eval(Main, :($(Library.save)($q, $data...; $options...)))
         catch e
--- a/test/loadsave.jl
+++ b/test/loadsave.jl
@@ -215,3 +215,46 @@ end
 @testset "Absent file" begin
     @test_throws SystemError load("nonexistent.oops")
 end
+
+# Stack overflows (issue #141)
+module IOrecursion1
+using FileIO
+# One shouldn't extend FileIO.load, but what happens if someone does anyway?
+FileIO.load(file::File{format"IOrecursion1"}) = 1
+end
+
+module IOrecursion2
+using FileIO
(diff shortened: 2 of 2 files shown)


timholy:
CC @simonster

timholy:
I should point out that this is a dangerous PR that I won't merge unless there's substantial buy-in from other folks. It will break JLD (and maybe many other packages) until they stop extending `load` and `save`:
```julia
julia> save("myfile.jld", Dict("x"=>5))
INFO: Recompiling stale cache file /home/tim/.julia/lib/v0.6/JLD.ji for module JLD.
Error encountered while saving "/tmp/myfile.jld".
Fatal error:
ERROR: JLD writer error: module should not extend FileIO.save
  Will try next writer.
```
Though through the magic of dispatch, it works if you just try a second time:
```julia
julia> save("myfile.jld", Dict("x"=>5))
```

ssfrr:
I think that if the FileIO API is that you shouldn't extend `load` and `save` but instead define your own, it seems reasonable to enforce that (especially if not doing so causes stack overflows).

If the affected package authors get a heads up they should be able to fix things on their side without breaking anything, before this gets merged. Another option could be to make it a warning at first, so there's a sort of deprecation period.

It would probably be good to include a specific note in the [Implementing Loaders/Savers](https://github.com/JuliaIO/FileIO.jl#implementing-loaderssavers) section of the README that says that your package should define `load` and `save` methods (but not export them).

timholy:
Thanks @ssfrr. See #164. If we decide to go ahead with this PR, before merging I propose that we send something out to discourse and give folks a couple of weeks to adjust.

For the record:
```julia
julia> Pkg.dependents("FileIO")
50-element Array{AbstractString,1}:
 "Luxor"                 
 "Kpax3"                 
 "Images"                
 "ELF"                   
 "OMETIFF"               
 "MFCC"                  
 "GLAbstraction"         
 "NRRD"                  
 "LibSndFile"            
 "RData"                 
 "AndorSIF"              
 "JLD"                   
 "LasIO"                 
 "GLBooks"               
 "PkgBenchmark"          
 "Weber"                 
 "Meshes"                
 "WAV"                   
 "StatFiles"             
 "GLVisualize"           
 "GLWindow"              
 "RigidBodyTreeInspector"
 "TestImages"            
 "MeshIO"                
 "BedgraphFiles"         
 "Ogg"                   
 "VideoIO"               
 "TensorFlow"            
 "ProfileView"           
 "FeatherFiles"          
 "MP3"                   
 "JLD2"                  
 "CSVFiles"              
 "ColorSchemes"          
 "Opus"                  
 "SwiftObjectStores"     
 "ReferenceTests"        
 "QuartzImageIO"         
 "MetaImageFormat"       
 "FLAC"                  
 "ImageView"             
 "ImageMagick"           
 "ExcelFiles"            
 "ImagineFormat"         
 "Netpbm"                
 "Augmentor"             
 "Caesar"                
 "GslibIO"               
 "EMIRT"                 
 "RDatasets"             
```

We could download all of these and see which ones seem to be extending FileIO's methods. This won't catch private repos, but would be better than nothing. I feel guilty that we let this go so long without documenting it unambiguously.

rapus95:
Stumbled over this, looked into the changes, found the ancient 2017 syntax. Happy to have newer better syntax.
Anyway, are there similar plans or was it just forgotton to be closed?

timholy:
#295 implemented a warning

PR #155, opened 2017-09-27
Custom function to find bedgraph magic bytes

CiaranOMara:
Bedgraph magic bytes were not guaranteed to be at the start of the file header. They could be at practically any location in a header of indeterminate length.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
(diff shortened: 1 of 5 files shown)


timholy:
Just an FYI, this hasn't been merged because it fails tests.

timholy:
(Click the "Details" or red x for more info.)

PR #157, opened 2017-11-06
WIP Register fst file format

davidanthoff:
Don't merge yet, we first need to sort out various things in the upstream packages.


timholy:
Are there really no magic bytes? In this case fst seems to mean several things:

- https://fileinfo.com/extension/fst
- http://whatis.techtarget.com/fileformat/FST-Linkable-program-dBFast
- https://github.com/fstpackage/fst

I bet you don't mean the first, but even knowing you I can't guess which of the other two you're hoping to support.

davidanthoff:
I’m trying to find out. This will hopefully support the third format.

davidanthoff:
Because we asked them to, they are actually adding a magic byte into the next version :) I'll follow up on this once everything is ready, will probably still be a while.

timholy:
Open source at its finest!

ViralBShah:
This is an old one - but is it ok to merge it now?

davidanthoff:
I have absolutely no idea :) I would close this PR, whatever it did is probably outdated?

PR #158, opened 2017-11-20
Custom function to find bedgraph magic bytes (take 2)

CiaranOMara:
Sorry @timholy, I somehow closed #155 when tidying my develop branch.

This pull request additionally converts the magic bytes to a string for the use of the `contains(::AbstractString, ::AbstractString)` method.

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -29,8 +29,6 @@ add_format(format"SAS", UInt8[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
     0xcf, 0xbd, 0x92, 0x08, 0x00, 0x09, 0xc7, 0x31, 0x8c, 0x18, 0x1f,
     0x10, 0x11], [".sas7bdat"], [:StatFiles, LOAD])
 
-add_format(format"bedGraph", UInt8[0x74, 0x79, 0x70, 0x65, 0x3D, 0x62, 0x65, 0x64, 0x47, 0x72, 0x61, 0x70, 0x68], [".bedgraph"], [:BedgraphFiles])
-
 # Image formats
 add_format(format"PBMBinary", b"P4", ".pbm", [:ImageMagick])
 add_format(format"PGMBinary", b"P5", ".pgm", [:Netpbm])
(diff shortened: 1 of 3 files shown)


src/loadsave.jl
```diff
@@ -3,9 +3,11 @@ const sym2saver  = Dict{Symbol,Vector{Symbol}}()
 
 function is_installed(pkg::Symbol)
     isdefined(Main, pkg) && isa(getfield(Main, pkg), Module) && return true # is a module defined in Main scope
-    path = Base.find_in_path(string(pkg)) # hacky way to determine if a Package is installed
-    path == nothing && return false
-    return isfile(path)
+    version = try Pkg.installed(string(pkg)) end # Note: If pkg is registered, but not installed, Pkg.installed() returns nothing. Also, try block returns nothing if something fails.
```
SimonDanisch: I'm pretty sure, that this was the initial version and we dumped it at some point because it was suboptimal. It was either very slow or probably not working in some cases - not 100% sure anymore :(

CiaranOMara: I had a quick flick through the tree. The is_installed method was first introduced by @timholy in commit 2df4c5a6 and the last change or current 3 lines were introduced by you in commit 64c7a728.

There were 3 other commits in between that alter the method in a relevant way:
- 7256f85e
- 0e98cd5e
- 0ec071bd

Prior to the last relevant change, it looks package installation was checked by checking a dictionary for symbols -- I'm too new to Julia to know the context.

Is it that the 3 lines in question are determining the difference between whether a package is registered and installed?

SimonDanisch: Hm maybe I only tried it out locally. Well, it is supposed to tell whether a package can be `used`. Maybe the best way is to mimic what base does when you do `using Module` - I think there should be some functions flying around that look in the load path etc.

SimonDanisch: https://github.com/JuliaLang/julia/blob/10ab2beeb09936ad2cf9d02fce808293e960a519/base/loading.jl#L90 this looks promising :)

SimonDanisch: Oh that isn't on 0.6... https://github.com/JuliaLang/julia/blob/v0.6.1/base/loading.jl#L102 so this uglier method is the one, I guess ;)

CiaranOMara: I admit the find_package method does look more direct.

CiaranOMara: Would you like me to roll back this pull request to 133b993?

timholy:
Thanks!

PR #159, opened 2017-11-22
Add support for Rdata single format (.rds files)

jsams:
In addition to the more common rda/RData files generated by R's `save` function, there is also the `saveRDS` function which saves a single data object rather than the named collection as in `save`. I am adding support for these objects to the RData package and wanted to register the file format here.

rds PR in RData: https://github.com/JuliaStats/RData.jl/pull/33


alyst:
Travis CI failure on nightly is due to `find_in_path()` being replaced by `find_package()`; that's not related to this PR.

timholy:
Overall this looks good. Just to check, though, from [here](https://stackoverflow.com/questions/21370132/r-data-formats-rdata-rda-rds-etc) I went to [this page](http://stat.ethz.ch/R-manual/R-devel/library/base/html/readRDS.html) and found the following line:

>  However, `save` writes a single line header (typically `"RDXs\n"`) before the serialization of a single object (a pairlist of all the objects to be saved). 

Does that apply here, or is that some other format?

alyst:
IIUC in .rdata, there's `RDX2\n` or `RDA2\n` header, followed by the repetition of `X\n2\n` or `A\n2\n`.
In .rds the header line is missing but `X\n2\n`/`A\n2\n` still remains, and it's read by `detect_rdata_single()`.

timholy:
Assuming that's an R-written file, then of course this is worth supporting. 2 characters is a pretty wimpy magic number, but that's not your problem. Together with the file extension it seems OK.

alyst:
Thanks!

PR #161, opened 2017-11-24
defines new load/save functions in Dummy module rather than extending FileIO's

ssfrr:
Loader and saver packages should define their own load/save functions, not
extend FileIO.load and FileIO.save. This fixes the `Dummy` module in the tests
to conform to this.


ssfrr:
failing on nightly, but I think that might be unrelated to this PR.

timholy:
Thanks!

PR #164, opened 2017-11-27
Clarify that load and save should not be extended

timholy:
Implementing the excellent suggestion made by @ssfrr in https://github.com/JuliaIO/FileIO.jl/pull/154#issuecomment-335932537. I hadn't realized that this wasn't clearly documented.


SimonDanisch:
Sounds good to me, since almost anyone who started extending FileIO run into this...!

PR #175, opened 2018-03-07
1st pass at fixing v0.7 deprecations

cormullion:
This PR is an attempt to fix most of the deprecations and errors that you get when running the FileIO tests on v0.7.

These are the only two left which I couldn't track down:

```
julia> Pkg.test("FileIO")
WARNING: Base.Pkg is deprecated, run `using Pkg` instead
 in module Main
[ Info: Testing FileIO
┌ Warning: Vector{UInt8}(s::String) will copy data in the future. To avoid copying, use `unsafe_wrap` or `codeunits` instead.
│   caller = top-level scope
└ @ Core :0
┌ Warning: Vector{UInt8}(s::String) will copy data in the future. To avoid copying, use `unsafe_wrap` or `codeunits` instead.
│   caller = top-level scope
└ @ Core :0

```

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
-julia 0.6
+julia 0.7.0-
 Compat 0.28.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,3 @@
 julia 0.7.0-
 Compat 0.28.0
+Nullables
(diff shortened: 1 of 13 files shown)


cormullion:
Well, it's not working yet, but perhaps it can serve as a basis for someone else to finish it off... :)

cormullion:
tests started failing again with latest Julia (0.7.0-DEV.4698 (2018-03-26 18:49 UTC) Commit a9eaa4f). I think it's to do with the file detection of unknown files. No idea how to fix.

PR #177, opened 2018-03-20
clarifies stream seeking for querying and detection functions

### LINKED ISSUE #163: should `load(::Stream, ...)` methods skip magic bytes?
In the README, the example for loading from a stream is:

```julia
function load(f::File{format"PNG"})
  open(f) do s
  skipmagic(s) # skip over the magic bytes
  # You can just call the method below...
  ret = load(s)
  # ...or implement everything here instead
  end
end

# You can support streams and add keywords:
function load(s::Stream{format"PNG"}; keywords...)
  # s is already positioned after the magic bytes
  # Do the stuff to read a PNG file
  chunklength = read(s, UInt32)
  ...
end
```

This works if the user calls `load("somefile.png")`, which will end up calling the `load(::File, ...)` method, which handles skipping the magic bytes before calling `load(::Stream)`. What I'm not super clear on is how this is intended to be used directly. e.g. I couldn't just do:
```julia
open("somefile.png") do io
    load(io)
end
```

In ImageMagick.jl it [looks like](https://github.com/JuliaIO/ImageMagick.jl/blob/78e573c541a63fac98e01909ca82ef0f3d093cc7/test/constructed_images.jl#L195-L201) `load` is called on the IO stream directly without skipping the magic(k) bytes.

So if it seems like the way ImageMagick does it (leaving the stream at the very beginning of the file when calling `load(::Stream)`) I can put in a documentation PR that clarifies that.

ssfrr:
The query infrastructure was already handling seeking around the stream
before and after calling the detect functions, so they shouldn't have to
worry about it. This also sets the example in the README to skip the magic
bytes in `load(::Stream)` instead of `load(::File)`.

fixes #163


timholy:
Ugh, I just noticed this. If you want something like this it will likely be easier to start from scratch. I am not 100% certain we've been as careful as we should be about the stream position upon exit, or even what the right default really is. Maybe the detector should leave the position as-is after successful detection?

Happy to help you get this over the finish line if you're still interested.

Issue #173, opened 2018-01-25
query metadata feature?


bjarthur:
`query` returns the file type (e.g. tiff), but i frequently need to know the size (e.g. 1024x768 pixels) without actually loading the whole image.  does FileIO provide an interface for this?

pretty straightforward to implement this for ImageMagick as follows:

```
function get_image_dimensions(file)
    wand = MagickWand()
    readimage(wand, file)
    xy=size(wand)
    z=getnumberimages(wand)
    z==1 ? xy : (xy...,z)
end
```

happy to submit a PR, just tell me how to do it.  thanks.

bjarthur:
i propose we had to FileIO a `metadata` function, to add to `load` and `save`, which for files containing images returns a sub-package (e.g. ImageMagick.jl) defined type that contains the width, height, depth, eltype, etc.  `METADATA` would be defined to add to the existing `LOAD` and `SAVE`.  similarly, `add_metadata` would complement `add_{loader,saver}`.  happy to work on a PR but would like to hear what people think first.

timholy:
Does one need a separate `add_metadata` or is that implicit with `add_loader`? Is there a case where the loading package *wouldn't* be the one to return the metadata?

PR #178, opened 2018-04-11
add metadata(), to avoid loading the whole file

bjarthur:
closes https://github.com/JuliaIO/FileIO.jl/issues/173

also see https://github.com/JuliaImages/Images.jl/pull/714 and https://github.com/JuliaIO/ImageMagick.jl/pull/111

### DIFF
--- a/src/FileIO.jl
+++ b/src/FileIO.jl
@@ -24,7 +24,8 @@ export DataFormat,
        savestreaming,
        skipmagic,
        stream,
-       unknown
+       unknown,
+       metadata
 
 import Base.showerror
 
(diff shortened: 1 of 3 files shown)


bjarthur:
@timholy could you please review this and the two linked PRs?  thanks.

bkamins:
@bjarthur, @timholy - I could not find a documentation of contract for `metadata` function defined in FileIO.jl.

In parallel we are considering adding `metadata` function to DataAPI.jl.

I have the following questions:
* would it make sense for `FileIO.metadata` to add a method to `DataAPI.metadata` instead of creating a new function;
* in this case would it make sense/be possible to return e.g. `NamedTuple` instead of `Tuple` (in DataAPI.jl a natural interface is to return an object having key => value pairs)

Or maybe there is some other solution to the name clash?

PR #179, opened 2018-05-15
Add support for fileio_* names

davidanthoff:
This changes how FileIO looks for load and save functions in packages that implement support for a given file format. With this PR, it will first look for functions named ``fileio_load``, ``fileio_save`` etc. If it finds these, it will use them, otherwise it will fall back to the old way of looking for a ``load`` and ``save`` function in the package. So this PR should be completely backwards compatible.

The main benefit of this PR is that it allows a package for a file format to re-export the ``load`` and ``save`` function from FileIO. For example, [CSVFiles.jl](https://github.com/davidanthoff/CSVFiles.jl) will be able to export the function ``FileIO.load`` and ``FileIO.save`` with this patch. Right now it cannot do that: it needs to define a ``CSVFiles.load`` and ``CSVFiles.save`` function, and once those are defined, ``CSVFiles`` can no longer export the ``FileIO`` version of these functions. With this patch, I can change the implementation in CSVFiles.jl to implement ``fileio_load`` and ``fileio_save`` functions, and that opens up the possibility to reexport the functions from FileIO from CSVFiles.

Why is this useful? Right now it is really annoying that one can't just do ``using CSVFiles`` and then start to work, one always has to do a ``using FileIO`` as well. That is just weird and not a good user experience.

I'll add some tests and README stuff if @timholy and @SimonDanisch sign off on the general approach here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,13 @@ Consequently, **packages should define "private" `load` and `save` methods (also
 `loadstreaming` and `savestreaming` if you implement them), and not extend
 (import) FileIO's**.
 
+If you run into a naming conflict with the `load` and `save` functions
+(for example, you already have another function in your package that has
+one of these names), you can instead name your loaders `fileio_load`, 
+`fileio_save` etc. Note that you cannot mix and match these styles: either
+all your loaders have to be named `load`, or all of them should be called
+`fileio_load`, but you cannot use both conventions in one module.
+
 `load(::File)` and `save(::File)` should close any streams
 they open.  (If you use the `do` syntax, this happens for you
 automatically even if the code inside the `do` scope throws an error.)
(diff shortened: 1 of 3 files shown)


SimonDanisch:
Sounds good to me!

davidanthoff:
@SimonDanisch I think @timholy is busy these days with other stuff. Should we just merge this and tag a release? It would allow me to finish some other projects that I have pending :)

davidanthoff:
> can you rebase to allow 0.7 ci failure?

Done.

BUT, I realize I promised tests and README updates if folks are ok with the general idea. So, let me add those and then we can merge.

SimonDanisch:
OK, ready when you are ;)

davidanthoff:
Oh, and I see that you just tagged 0.8.0... Maybe we could even include this here in 0.8.0, before it gets merged in METADATA? Or tag a 0.9.0 on its heels? Sorry for the extra work here...

SimonDanisch:
Let's just tag 0.9 ;) No need to be greedy on tags!

Issue #167, opened 2017-12-17
FileIO (master) does not precompile on julia nightly (master)


bluesmoon:
```
julia> VERSION
v"0.7.0-DEV.3018"

julia> Pkg.status("FileIO")
 - FileIO                        0.6.1+             master
```

```
ERROR: LoadError: LoadError: TypeError: in Type{...} expression, expected UnionAll, got typeof(Base.Nullable)
Stacktrace:
 [1] top-level scope
 [2] include at ./boot.jl:283 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:503
 [4] include at ./sysimg.jl:15 [inlined]
 [5] include(::String) at /root/.julia/v0.7/FileIO/src/FileIO.jl:3
 [6] top-level scope
 [7] include at ./boot.jl:283 [inlined]
 [8] include_relative(::Module, ::String) at ./loading.jl:503
 [9] include(::Module, ::String) at ./sysimg.jl:15
 [10] top-level scope
 [11] eval at ./boot.jl:286 [inlined]
 [12] top-level scope at ./<missing>:2
in expression starting at /root/.julia/v0.7/FileIO/src/query.jl:242
in expression starting at /root/.julia/v0.7/FileIO/src/FileIO.jl:29
ERROR: Failed to precompile FileIO to /root/.julia/lib/v0.7/FileIO.ji.
```

bluesmoon:
This affects the precompile of `DataFrames`

johannspies:
I get the same on 
```
Julia Version 0.7.0-DEV.5154
Commit 47edaef883* (2018-05-22 02:22 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, broadwell)

 Pkg.status("FileIO")
 - FileIO                        0.7.0+             master
```


```
ERROR: LoadError: LoadError: TypeError: in Type{...} expression, expected UnionAll, got typeof(Base.Nullable)
Stacktrace:
 [1] top-level scope
 [2] include at ./boot.jl:314 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1069
 [4] include(::Module, ::String) at ./sysimg.jl:29
[...]

PR #181, opened 2018-05-22
fixes for 0.7

SimonDanisch:


SimonDanisch:
Should fix: https://github.com/JuliaIO/FileIO.jl/issues/167

SimonDanisch:
Api for filename(::Stream) is changed to use `Union{Nothing, String}`. Hope this is not too breaking!

SimonDanisch:
Packages shouldn't really use `filename` in most cases. At least ImageMagick, the one were I most expect it, isn't using it :)

SimonDanisch:
@timholy @davidanthoff @ssfrr any concerns about using `filename(::Stream)::Union{Void, String}` instead of `filename(::Stream)::Nullable{String}`?

ssfrr:
That seems to be the direction things are going, so no problem here.

davidanthoff:
Sounds good to me!

PR #187, opened 2018-07-03
Fredrikekre fe/0.7

SimonDanisch:
@fredrikekre fixes for 0.7 plus one more fix!
This passes locally on windows with the newest master build of julia, so let's merge if there are no new failures!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ os:
   - linux
   - osx
 julia:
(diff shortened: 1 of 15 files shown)


SimonDanisch:
Well, that does look better, appveyor is still failing for mysterious reasons, but I guess my local windows test run will need to do the trick here.

Issue #189, opened 2018-07-04
RFC: Potential redesign of how loaders/savers register with FileIO


KristofferC:
FileIO right now does quite a lot of "unorthodox" things which makes it a bit troublesome to upgrade to the new code loading, package management system. Examples of these are evaluating methods into other packages, reflecting on method tables at runtime etc.  In addition to being brittle, I believe these things heavily contribute to FileIO being slow to load and have quite complex code.

I suggest the following redesign for FileIO:

Packages that provide loading / saving functionality make themselves available to FileIO using something like:

```
add_loader(FileIO.Format{:PNG}, :ImageMagic, f_load)
add_loader_streaming(FileIO.Format{:PNG},  :ImageMagic, f_load_stream)
```

where `f_load` is a function defined in ImageMagic.jl that loads a file and `f_load_stream` is a function that loads from a buffer. Same applies for savers.

The crucial difference from before is that packages tell FileIO that they are available, FileIO itself doesn't try to find available loaders by looping through method tables and injecting load functions into packages etc.

Internally, FileIO would have a collection of loaders (and savers etc)

```jl
const loaders = Dict{FileIO.Format, Vector{Loader}}()
```

where, conceptually, a `Loader` is a

```jl
struct Loader
    package::Symbol
    load_function #::Function
end
```

The loader function `add_loader` would then look something like

```jl
function add_loader(t::Type{FileIO.Format{T}}, pkg::Symbol, f) where {T}
    haskey(loaders, t) || loaders[t] = Vector{Loader}()
    push!(loaders, Loader(pkg, f))
end
```

FileIO also has a registry which has the following functionalities

* From a file, return what file type the file is.
* From a file type, provide a list of possible packages that can load / save that file type.
* Also provide a priority list in case multiple loaders / savers are available.

This is already provided in `registry.jl` and it could likely be kept with few changes. The changes I can see is that the packages cannot be identified by just the name anymore so should use a UUID as well. 

Upon doing `FileIO.load("xxx.yyy")` the following steps are taken.

1. Use the registry to determine the file type (based on extension or magic bytes), say `FileIO.Format{T}`.
2. Based on the file type, look up pontential loaders in the `loaders` dictionary (`loaders[FileIO.Format{T}]`)
3. If there are multiple loaders, determine which one based on the priority list in the registry.
4. When a loader is determined, call the registered loading function for that loader.
5. If no loader was found, display an error message and give a suggestion of possible package that can load the file.

[...]

PR #190, opened 2018-07-04
Workaround for hot-loading handlers in Julia v0.7

RalphAS:
While we're waiting for implementations of a redesign (#189) here is a workaround for the current approach.


SimonDanisch:
Trying this out right now! I guess this removes the functionality of installing packages automatically, which seems ok to me.

SimonDanisch:
I can't really test all relevant cases (E.g. ImageMagick, even your PR, fails right now), but looks like it works fine! Thanks for doing this!
@timholy should we just hit merge?

PR #191, opened 2018-07-18
add ProteinDataBank format

Mikhail-Vaganov:
As for now this file format will be used for export of simulation data calculated using tools from DIffEqPhysics

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -283,4 +283,4 @@ add_format(format"MetaImage", "ObjectType", ".mhd", [:MetaImageFormat])
 
 add_format(format"vegalite", (), [".vegalite"], [:VegaLite])
 
-add_format(format"ProteinDataBank", (), ".pdb")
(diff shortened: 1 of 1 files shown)


SimonDanisch:
You'll need to add the package that supports loading/saving!
E.g.:
```Julia
add_format(format"ProteinDataBank", (), [".pdb"], [:YourPackageName]) # for loading and saving
add_format(format"ProteinDataBank", (), [".pdb"], [:YourPackageName, LOAD]) # for only loading (SAVE for saving)
```

Mikhail-Vaganov:
Sure. I thought about that, but suggested, that it is for some specific formats. Thanks!

timholy:
Does the DiffEqPhysics package actually load or save this format? This seems like an unusual choice, most file formats have a dedicated IO package separate from any particular computational task.

timholy:
This isn't mergeable without changes. Closing as stale.

PR #196, opened 2018-08-07
`info` should not extend `Base.info`

JeffBezanson:
This was always rather different functionality, and Base.info has been removed. Fixes tests on 1.0.


JeffBezanson:
Oops, I see Andreas beat me to this.

Issue #208, opened 2018-10-26
Fatal error: warn not defined


cormullion:
I think this is an error,  where it perhaps should be a warning. 

```
All errors:
   UndefVarError(:warn)
   ArgumentError("Package ImageMagick not found in current path:\n- Run 
 `import Pkg; Pkg.add(\"ImageMagick\")` to install the ImageMagick package.\n")
Fatal error:
ERROR: UndefVarError: warn not defined
Stacktrace:
 [1] require(::Module, ::Symbol) at ./loading.jl:820
 [2] eval at ./boot.jl:319 [inlined]
 [3] topimport(::Symbol) at /Users/me/.julia/packages/FileIO/lt9I0/src/loadsave.jl:13
 [4] checked_import(::Symbol) at /Users/me/.julia/packages/FileIO/lt9I0/src/loadsave.jl:30
 [5] #load#27(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function,
 ::FileIO.File{FileIO.DataFormat{:PNG}}) at /Users/me/.julia/packages/FileIO/lt9I0/src/loadsave.jl:180
```
Should this use `@warn`? (The FileIO code is a bit too tangled for me to follow.)

FileIO v1.0.2, Julia Version 1.0.1

SimonDanisch:
if there is a `warn` somewhere, it should definitely use `@warn`!

cormullion:
Looks like this one? 

https://github.com/JuliaIO/FileIO.jl/blob/fb403010ff0ad42c518150c2aac074ab51e58301/src/error_handling.jl#L52

I'll have a go at fixing it.

cormullion:
Huh, I didn't get far. :(

I changed the two lines with `warn` to this:

```julia
"""
Handles error as soon as they get thrown while doing IO
"""
function handle_current_error(e, library, islast::Bool)
    bt = catch_backtrace()
    bts = sprint(io->Base.show_backtrace(io, bt))
    message = islast ? "" : "\nTrying next loading library! Please report this 
issue on the Github page for $library"
    @warn string(e, bts, message)
end
handle_current_error(e::NotInstalledError) = @warn string("lib ", e.library, " 
not installed, trying next library")
```

But I still get this error:

```
julia-1.0> using FileIO; load("nfound.png")
Errors encountered while loading "nfound.png".
All errors:
   UndefVarError(:warn)
[...]

PR #209, opened 2018-10-27
replace warn with @warn

cormullion:
Replace `warn` with `@warn`. (fixes https://github.com/JuliaIO/FileIO.jl/issues/208)


SimonDanisch:
the test failures must be unrelated!  thanks for the fix!

Issue #213, opened 2018-11-06
rethrow(exc) not allowed outside a catch block


staticfloat:
Some errors kill `FileIO` because it tries to `rethrow()` from outside of a `catch` block.

```
Stacktrace:
 [1] rethrow(::ErrorException) at ./error.jl:52
 [2] handle_error(::ErrorException, ::FileIO.File{FileIO.DataFormat{:JPEG}}) at /home/sabae/.julia/packages/FileIO/lt9I0/src/error_handling.jl:80
 [3] handle_exceptions(::Array{Any,1}, ::String) at /home/sabae/.julia/packages/FileIO/lt9I0/src/error_handling.jl:75
 [4] #load#27(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::FileIO.File{FileIO.DataFormat{:JPEG}}) at /home/sabae/.julia/packages/FileIO/lt9I0/src/loadsave.jl:193
 [5] load(::FileIO.File{FileIO.DataFormat{:JPEG}}) at /home/sabae/.julia/packages/FileIO/lt9I0/src/loadsave.jl:172
 [6] #load#13(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::String) at /home/sabae/.julia/packages/FileIO/lt9I0/src/loadsave.jl:118
 [7] load at /home/sabae/.julia/packages/FileIO/lt9I0/src/loadsave.jl:118 [inlined]
 [8] load_img(::String) at /home/sabae/.julia/dev/Metalhead/src/preprocessing.jl:13
```

musm:
Do you my any chance have a MWE?

musm:
I wonder if @c42f has any comments since he implemented the new exception system.

Glancing at the error handling file, indeed there seems to be some issues I believe with the logic
https://github.com/JuliaIO/FileIO.jl/blob/master/src/error_handling.jl#L80

c42f:
Using `rethrow()` as is done in `handle_error()` probably never worked properly because the exception can get mixed up with the wrong backtrace. A simple fix is to simply `throw(e)` rather than using `rethrow(e)`. This won't give a useful backtrace either but at least it will be consistent.

Better though to capture the each backtrace along with the exception using `catch_backtrace()` and communicate the set of those errors to the user somehow. There's a mixture of logging and throwing already in the package...

PR #215, opened 2018-11-19
Stopgap fix of rethrow

musm:
Courtesy @c42f  (thanks)

A  more indepth fix should be pursued in follow up PR's, this at least fixes the problem reported in https://github.com/JuliaIO/FileIO.jl/issues/213


SimonDanisch:
Thank you!

PR #221, opened 2019-02-07
Adding .las format to registry

leytzher:
Hi, I am proposing an addition to the registry. This addition is for the LAS file format which is a text based format without magic bytes.

The Log ASCII Standard (LAS) files are used for borehole data such as geophysical, geological, or petrophysical logs in oil and gas wells.
The LAS file specification has been published by [the Canadian Well Logging Society] (http://www.cwls.org/las/)

Note this is not for LiDAR data (also called "LAS files").


leytzher:
I made a mistake on my pull request. 
I will fix it and resubmit. Thanks!

PR #223, opened 2019-03-13
Don't export `info`

rofinn:
The implementation of `info` doesn't even match the meaning of what `Base.info` was.


SimonDanisch:
That makes sesne^^ Thanks ;)

PR #230, opened 2019-07-03
break up query.jl into seperate files

oxinabox:
I was having trouble finding things in the package.
In particular I was looking for the `Stream` type, and couldn't find it.

I think this is a bit of a cleaner restructuring of the files.

No features have been added or removed it is just code being moved around.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *.jl.cov
 *.jl.mem
 src/all_magic_bytes/
+Manifest.toml
+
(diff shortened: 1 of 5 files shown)


Tokazama:
I agree that this is a nicer way of organizing things.

oxinabox:
Bump

SimonDanisch:
Thank you ;) Looks good to me! @timholy, what do you think ?

SimonDanisch:
Thank you :)

PR #237, opened 2019-11-21
Dummy module now defines its own load/save instead of extending

ssfrr:
Some of these changes were made in #161 but it looks like I accidentally reverted them in #78. Possibly a bad merge.

This also adds tests for keyword arguments in `load` and `save`, making sure they get passed through to the handling package.


ssfrr:
Thanks for the quick review!

PR #239, opened 2019-11-21
bumps the minor version number to 1.1.0

ssfrr:


timholy:
Seems like this has already been done & released.

PR #247, opened 2019-12-21
Use recommended invokelatest function instead of eval

vtjnash:
Using `Core.eval` is slow and often full of minefields. The `invokelatest` API is what's recommended for this purpose to avoid those.


timholy:
Fails tests.

PR #250, opened 2020-01-01
Add support for HTML saving via MIME types

davidanthoff:
This will enable saving of vega and vega-lite files to HTML.

I've added a custom MIME type called `application/vnd.julia.fileio.htmlfile` for situations where a type does not want to provide a `show` method for `text/html`, but still wants to support saving as HTML files (this is a case I need to support in VegaLite.jl).


timholy:
Willing to merge this in principle (this is an area I don't know a lot about), but the tests are currently failing. Does it just need to be retriggered or is there something missing here?

davidanthoff:
Ok, tests are fixed now!

davidanthoff:
Thanks! Could you also tag a release?

timholy:
https://github.com/JuliaRegistries/General/pull/9078

PR #254, opened 2020-02-06
Register JLPROF format for storing Julia profiling data

timholy:
This defines a new format for saving profiling data in a robust manner. See the companion PR: https://github.com/timholy/FlameGraphs.jl/pull/12.


timholy:
Tests pass. I'll give folks a couple of days to object, then merge.

PR #257, opened 2020-02-15
Using PNGFiles.jl via ImageIO.jl (lazy loader) for .png image files

ianshmean:
https://github.com/JuliaIO/PNG.jl isn't yet registered, but is functional with this branch.


ianshmean:
As @davidanthoff pointed out on slack, we need to make  sure the MIME for png isn't messed  up by this. Just making a note as a reminder

timholy:
Need to wait for https://github.com/JuliaRegistries/General/pull/9508 (3 days).

johnnychen94:
Does this mean that to load a png file we'll need to install `PNG` instead of `ImageMagick`? If so then a lot of packages' test codes will be broken without adding `PNG` to the `Project.toml`'s `extra` section.

ianshmean:
> Need to wait for JuliaRegistries/General#9508 (3 days).

Indeed, and as @johnnychen94's point illustrates, even once that's registered this needs careful thinking before merging

ianshmean:
It makes me think that if we were to move to this approach, it might be worth waiting until  the big three, PNG.jl, JPEG.jl and TIFF.jl, were ready to make the switchover easier?

timholy:
IIRC packages get precedence by argument order. So we could leave QuartzImageIO and ImageMagick as valid loaders but put PNG as the first choice.

johnnychen94:
If the final goal is to deprecate ImageMagick(is it?), then I think the following procedure makes sense:

1. add a deprecation warning in `ImageMagick` and `QuartzImageIO` for specific image format to encourage users to try out `PNG.jl`
2. leave `ImageMagick` and `QuartzImageIO` as fallbacks in `FileIO` for a period
3. unregister `ImageMagick` and `QuartzImageIO` from `FileIO`

timholy:
I think we're likely to have ImageMagick around for some time for less commonly-used formats. Just restoring those lines to the `add_format` seems like it should be enough even without changing any of the other packages.

src/registry_setup.jl
```diff
@@ -81,7 +81,7 @@ end
 `add_format(fmt, magic, extention)` registers a new `DataFormat`.
 For example:
 
-    add_format(format"PNG", (UInt8[0x4d,0x4d,0x00,0x2b], UInt8[0x49,0x49,0x2a,0x00]), [".tiff", ".tif"])
+    add_format(format"TIFF", (UInt8[0x4d,0x4d,0x00,0x2b], UInt8[0x49,0x49,0x2a,0x00]), [".tiff", ".tif"])
```
ianshmean: Just wanted to highlight that I thought this was a typo, so corrected it

davidanthoff:
Could you try whether this would break the existing MIME based PNG story? Here is a simple test:
```
using VegaLite, VegaDatasets

dataset("cars") |> @vlplot(:point, x=:Acceleration) |> save("foo.png")
```

ianshmean:
@davidanthoff does this look right?
foo

davidanthoff:
Yes, if it produces a proper graph, all is good :)

timholy:
Just strip the WIP out of the title if you're ready for this to be merged.

ianshmean:
Will do. Need to wait for ImageIO to be released and a plan to be settled on for the test failures in https://github.com/JuliaIO/PNGFiles.jl/pull/5

ianshmean:
@timholy Ready to merge, given https://github.com/JuliaIO/PNGFiles.jl/pull/7 has been merged

timholy:
Oh boy oh boy oh boy!

PR #260, opened 2020-04-29
Add JLSO support

### LINKED ISSUE #234: add Support JLSO.jl
See https://github.com/invenia/JLSO.jl

rofinn:
Closes #234

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -6,7 +6,7 @@ add_format(format"JLD", (unsafe_wrap(Vector{UInt8}, "Julia data file (HDF5), ver
 add_format(format"JLD2", "Julia data file (HDF5), version 0.2", ".jld2", [:JLD2])
 add_format(format"GZIP", [0x1f, 0x8b], ".gz", [:Libz])
 add_format(format"BSON",(),".bson", [:BSON])
-add_format(format"JLSO",(),".jlso", [:JLSO])
+add_format(format"JLSO", (), ".jlso", [:JLSO])
 
 # test for RD?n magic sequence at the beginning of R data input stream
 function detect_rdata(io)
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -6,6 +6,7 @@ add_format(format"JLD", (unsafe_wrap(Vector{UInt8}, "Julia data file (HDF5), ver
 add_format(format"JLD2", "Julia data file (HDF5), version 0.2", ".jld2", [:JLD2])
 add_format(format"GZIP", [0x1f, 0x8b], ".gz", [:Libz])
 add_format(format"BSON",(),".bson", [:BSON])
+add_format(format"JLSO",(),".jlso", [:JLSO])
 
 # test for RD?n magic sequence at the beginning of R data input stream
 function detect_rdata(io)


src/registry.jl
```diff
@@ -6,6 +6,7 @@ add_format(format"JLD", (unsafe_wrap(Vector{UInt8}, "Julia data file (HDF5), ver
 add_format(format"JLD2", "Julia data file (HDF5), version 0.2", ".jld2", [:JLD2])
 add_format(format"GZIP", [0x1f, 0x8b], ".gz", [:Libz])
 add_format(format"BSON",(),".bson", [:BSON])
+add_format(format"JLSO",(),".jlso", [:JLSO])
```
omus, line 9:
```suggestion
add_format(format"JLSO", (), ".jlso", [:JLSO])
```

timholy:
Sorry to chime in here late. Can't JLSO possibly have magic bytes? Maybe I'm an outlier, but to me a file format that lacks magic bytes is seriously asking for trouble. There are so many examples where two formats choose the same extension.

rofinn:
> Sorry to chime in here late. Can't JLSO possibly have magic bytes? 

Not really, as it's just a wrapper format for a BSON file. I suppose we could choose to prepend some random magic bytes to each file, but then that would eliminate the ability for someone to inspect the file with any bson library regardless of language.

timholy:
Fair enough. If this does become an issue perhaps each file could start with a characteristic variable, e.g., the equivalent of `__format__ = "JLSO"` or something...

I am thinking of posting a rant on the topic (not directed in any way at you, this must be the 30th time this has come up) somewhere a little more public, but here's a dry run...

`<rant>`

The fact that people who design file formats don't seem to think about magic bytes is one of the more inane phenomena in all of computing. Suppose I have a collection of Float64s and Int64s that I want to save to the same file. Unless the order is prescribed, I'm going to need a mechanism for the file-reader to know what type is coming next. Suppose I told you that my strategy was to require you to name the file `FFIFIFFIIIFFF.dat`, where the sequence of letters in the filename indicates the sequence of Float64 or Int64. You'd probably think that's quite dumb, right? Because the user might want to name the file something else (`sensor_data_run27.dat` or something), or might have 3 different files in the same directory with the same sequence of types, etc.

People who design file formats would never do that, because they know there's another way to solve the problem: you "box" the variables by pre-pending them with a short sequence that encodes the type of the upcoming variable. Dumping raw bits into the file is a huge no-no unless the order of types is strictly determined by the format itself. You separate how the *user* wants to interact with the file (what s/he wants to use as a name) from essential information about how the *machine* needs to interact with the file.

Since file-format developers are accustomed to using type-tags for the individual chunks of data in them, why on earth don't they think one is needed **for the file as a whole**? Suppose someone hands me a folder full of files from a previous graduate student and asks me to continue the analysis. I see a `.gbv` file; is that Genie Timeline file or a PCB CAD file? Is that `.fst` file an audio file, a puzzle game file, or an R serialized dataframe file?

For every file out there that lacks magic bytes, the interpretation of that file might make sense in the head of possibly just one person, the person who saved it with that name. Once that person moves on, poof, that bit of ephemeral information is lost, and the file risks becoming meaningless. Aren't files supposed to have some permanence?

The epic failure to use magic bytes is all the more ironic because it's a fail by the very people who should know better. They would never make the same mistake in their handling of data *within* the format, yet there is no consideration given to how the format as a whole can be recognized. This insanity must stop. Use magic bytes, people.
 
`</rant>`

timholy:
Again, not directed at you. I'm quite excited about JLSO and just gave it a star, thanks so much for developing this and sharing it!

rofinn:
No worries, this does seem pretty rant worthy issue :) If we start seeing some overlap I'm happy to modify the JLSO format accordingly (even if that breaks direct BSON loading slightly). I was surprised that BSON didn't have magic bytes when I was adding this.

Awesome, thanks for taking a look at it! Hopefully, other folks in the community will find it useful.

PR #262, opened 2020-06-29
Add support for GPX.jl

alhirzel:
This is pending registration of GPX.jl and KMLTracks.jl and the merging of https://github.com/scls19fr/GPX.jl/pull/

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -358,4 +358,3 @@ add_format(format"FCS", "FCS", [".fcs"], [:FCSFiles])
 add_format(format"HTML", (), [".html", ".htm"], [:MimeWriter, SAVE])
 
 add_format(format"GPX", (), [".gpx"], [:GPX, LOAD])
-add_format(format"KML", (), [".kml"], [:KMLTracks, LOAD])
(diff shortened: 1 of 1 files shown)


scls19fr:
As mentioned in https://github.com/scls19fr/KMLTracks.jl/issues/6#issuecomment-651080628 maybe we should only focus on GPX.jl first (with both load and save and also with tests as mentioned in https://github.com/scls19fr/GPX.jl/pull/5#issuecomment-650943886

timholy:
Any magic bytes available? See my [rant](https://github.com/JuliaIO/FileIO.jl/pull/260#issuecomment-621766656).

alhirzel:
Hey @timholy, GPX is an XML-based format which does not have magic bytes in the traditional sense. The XML namespace is the closest thing. [See example.](https://en.wikipedia.org/wiki/GPS_Exchange_Format#Sample_GPX_document)

Perhaps there could be some content-aware file type detection mechanism which specializes for XML, but I feel it might end up being pretty fragile due to e.g. white space, file encoding, updated xmlns versions/URLs, etc. What are your thoughts? As @scls19fr mentioned we have a few more steps prior to taking this "prime time".

timholy:
Sorry for the delay. I'm willing to merge this if you're still interested.

alhirzel:
No problem Tim :)  I haven't used this for a little while, but this should still be good to merge.

timholy:
Thanks!

timholy:
Oh, GPX was never registered! I'm reverting #299. Sorry.

Issue #263, opened 2020-06-30
load() doesn't check whether file exists before offering alternative IO packages


ianshmean:
Here `fake/fake.png` doesn't exist. Instead of detecting that the file doesn't exist, FileIO returns file IO package options for the file format `.png` that aren't installed.

I believe `load` should always be used with files that exist.. so perhaps the first step in load should be checking for `isfile()`?

```
julia> using FileIO

julia> FileIO.load("fake/fake.png")
Errors encountered while loading "fake/fake.png".
All errors:
===========================================
Failed to open fake/fake.png
===========================================
ArgumentError: Package QuartzImageIO not found in current path:
- Run `import Pkg; Pkg.add("QuartzImageIO")` to install the QuartzImageIO package.

===========================================
ArgumentError: Package ImageMagick not found in current path:
- Run `import Pkg; Pkg.add("ImageMagick")` to install the ImageMagick package.

===========================================

Fatal error:
ERROR: Failed to open fake/fake.png
....
```

PR #265, opened 2020-07-13
Check for path errors eagerly during load/save (#263)

ianshmean:
This adds more eager checking for path errors during load & save for clearer error messages.
Fixes https://github.com/JuliaIO/FileIO.jl/issues/263

---

## Loading an a nonexistant file

Before:
```
julia> FileIO.load("fake/fake.png")
Errors encountered while loading "fake/fake.png".
All errors:
===========================================
Failed to open fake/fake.png
===========================================
ArgumentError: Package QuartzImageIO not found in current path:
- Run `import Pkg; Pkg.add("QuartzImageIO")` to install the QuartzImageIO package.

===========================================
ArgumentError: Package ImageMagick not found in current path:
- Run `import Pkg; Pkg.add("ImageMagick")` to install the ImageMagick package.

===========================================

Fatal error:
ERROR: Failed to open fake/fake.png
...
```

This PR:
```
julia> FileIO.load("fake/fake.png")
ERROR: ArgumentError: No file exists at given path: fake/fake.png
Stacktrace:
...
```

---

## Trying to save into a directory that doesn't exist

Before:
```
julia> FileIO.save("fake_dir/image.png", rand(UInt8,10,10))
Errors encountered while saving "fake_dir/image.png".
All errors:
===========================================
Could not open fake_dir/image.png for writing
===========================================
ArgumentError: Package QuartzImageIO not found in current path:
- Run `import Pkg; Pkg.add("QuartzImageIO")` to install the QuartzImageIO package.

===========================================
ArgumentError: Package ImageMagick not found in current path:
- Run `import Pkg; Pkg.add("ImageMagick")` to install the ImageMagick package.

===========================================
ArgumentError: Argument does not support conversion to png.
===========================================

Fatal error:
ERROR: Could not open fake_dir/image.png for writing
Stacktrace:
...
```

This PR:
```
julia> FileIO.save("fake_dir/image.png", rand(UInt8,10,10))
ERROR: ArgumentError: Directory of file path does not exist: fake_dir/image.png
Stacktrace:
...
```
UPDATE: Now `mkpath()` is run on the parent directory, if it doesn't exist and no error is thrown

---

## Trying to save to a path that's a directory

Before
```
julia> FileIO.save(format"PNG", "existing_dir", rand(UInt8,10,10))
Errors encountered while saving "existing_dir".
All errors:
===========================================
Could not open existing_dir for writing
===========================================
ArgumentError: Package QuartzImageIO not found in current path:
- Run `import Pkg; Pkg.add("QuartzImageIO")` to install the QuartzImageIO package.

===========================================
ArgumentError: Package ImageMagick not found in current path:
- Run `import Pkg; Pkg.add("ImageMagick")` to install the ImageMagick package.

===========================================
ArgumentError: Argument does not support conversion to png.
===========================================

Fatal error:
ERROR: Could not open existing_dir for writing
Stacktrace:
...
```

This PR:
```
julia> FileIO.save(format"PNG", "existing_dir", rand(UInt8,10,10))
ERROR: ArgumentError: Given file path is a directory: existing_dir
Stacktrace:
...
```

cc. @Johnnychen94 @timholy


ianshmean:
By the way, is `ArgumentError` the best option to throw here?

ianshmean:
Perhaps a solution could be to always `mkpath()` before passing to the backends? It's been mentioned before that that would be nice default behavior.

johnnychen94:
Not sure how others think, mkpath by default sounds a good solution to me

Octogonapus:
> Perhaps a solution could be to always `mkpath()` before passing to the backends? It's been mentioned before that that would be nice default behavior.

I would also like this to be the default.

ianshmean:
Ok, great. I've made that the case now

ianshmean:
Should we wait for more reviews?

timholy:
I think it's possible we'll be troubled by the path-creation, but this is undeniably better and might be perfect. Let's try it and see how we like it.

I think this merits a minor version bump, though? As a new feature? (That won't be breaking as we're over 1.0.)

ianshmean:
Minor bump sounds right. Thanks!

timholy:
Thanks for fixing this, BTW!

PR #271, opened 2020-08-13
Avoid creating global variables

timholy:
This prevents a scope warning on Julia 1.5 when running the tests


kimikage:
I think we just need to kick the two variables in "test/loadsave.jl" out of the `try` block.
https://github.com/JuliaIO/FileIO.jl/blob/1d498aee1e08276fe1e06c7f133e6ba0758b5666/test/loadsave.jl#L21-L25

Also, I prefer the following style.
```julia
@testset "Query" begin
    include("query.jl")
end
@testset "Load/Save" begin
    include("loadsave.jl")
end
@testset "Error Handling" begin
    include("error_handling.jl")
end
@testset "MIME" begin
    include("test_mimesave.jl")
end
```

PR #280, opened 2020-11-11
Replaced ImageMagick by Netpbm for load/save of Netpbm files.

magister-ludi:
Netpbm.jl has been extended to load and save all netpbm formats. The changes in this PR register Netpbm.jl to perform all load/save operations for these formats.

The default (i.e. without explicit format notation) save operations are the binary formats.

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -44,12 +44,12 @@ add_format(format"SAS", UInt8[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
 add_format(format"Parquet", "PAR1", [".parquet"], [:ParquetFiles, LOAD])
 
 # Image formats
-add_format(format"PBMBinary", b"P4", ".pbm", [:ImageIO])
-add_format(format"PGMBinary", b"P5", ".pgm", [:ImageIO])
-add_format(format"PPMBinary", b"P6", ".ppm", [:ImageIO])
-add_format(format"PBMText",   b"P1", ".pbm", [:ImageIO])
-add_format(format"PGMText",   b"P2", ".pgm", [:ImageIO])
-add_format(format"PPMText",   b"P3", ".ppm", [:ImageIO])
+add_format(format"PBMBinary", b"P4", ".pbm", [:ImageIO], [:Netpbm], [:ImageMagick])
+add_format(format"PGMBinary", b"P5", ".pgm", [:ImageIO], [:Netpbm])
+add_format(format"PPMBinary", b"P6", ".ppm", [:ImageIO], [:Netpbm])
+add_format(format"PBMText",   b"P1", ".pbm", [:ImageIO], [:Netpbm], [:ImageMagick, LOAD])
+add_format(format"PGMText",   b"P2", ".pgm", [:ImageIO], [:Netpbm], [:ImageMagick, LOAD])
+add_format(format"PPMText",   b"P3", ".ppm", [:ImageIO], [:Netpbm], [:ImageMagick, LOAD])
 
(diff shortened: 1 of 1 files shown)


ianshmean:
Just a thought. Might it be worth putting this into [ImageIO](https://github.com/JuliaIO/ImageIO.jl), beside PNGFiles and the other image loaders that will be added?

That way FileIO will be told to use ImageIO for Netpbm files, and ImageIO will direct to Netpbm.jl ?

There shouldn't be any noticeable overhead. It would just make it easier for users to install image readers in one go. There's even talk of adding `ImageIO.jl` to `Images.jl`

magister-ludi:
That's a sensible suggestion. I'll have a closer look at ImageIO and see what I can do.

magister-ludi:
I have submitted a PR (https://github.com/JuliaIO/ImageIO.jl/pull/13) to ImageIO that handles Netpbm formats (via Netpbm.jl). If that is merged, I can modify this PR so that FileIO registers ImageIO instead of Netpbm. If somebody else does the modification, that's not a problem for me.

magister-ludi:
I've updated the registry codeso that ImageIO is used to load and save Netpbm formats.

@ianshmean

src/registry.jl
```diff
@@ -44,12 +44,12 @@ add_format(format"SAS", UInt8[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
 add_format(format"Parquet", "PAR1", [".parquet"], [:ParquetFiles, LOAD])
 
 # Image formats
-add_format(format"PBMBinary", b"P4", ".pbm", [:ImageMagick])
-add_format(format"PGMBinary", b"P5", ".pgm", [:Netpbm])
-add_format(format"PPMBinary", b"P6", ".ppm", [:Netpbm])
-add_format(format"PBMText",   b"P1", ".pbm", [:ImageMagick, LOAD])
-add_format(format"PGMText",   b"P2", ".pgm", [:ImageMagick, LOAD])
-add_format(format"PPMText",   b"P3", ".ppm", [:ImageMagick, LOAD])
+add_format(format"PBMBinary", b"P4", ".pbm", [:ImageIO])
```
johnnychen94, line 47: Let's still keep ImageMagick, otherwise, it's a breaking change. Similar to other entries.

```suggestion
add_format(format"PBMBinary", b"P4", ".pbm", [:ImageIO], [:ImageMagick])
```

magister-ludi: The current version on master has the following entries:

- `P4 [:ImageMagick]`
- `P5 [:Netpbm]`
- `P6 [:Netpbm]`
- `P1 [:ImageMagick, LOAD]`
- `P2 [:ImageMagick, LOAD]`
- `P3 [:ImageMagick, LOAD]`

I can change it to

- `P4 [:ImageIO], [:Netpbm], [:ImageMagick]`
- `P5 [:ImageIO], [:Netpbm]`
- `P6 [:ImageIO], [:Netpbm]`
- `P1 [:ImageIO], [:Netpbm], [:ImageMagick, LOAD]`
- `P2 [:ImageIO], [:Netpbm], [:ImageMagick, LOAD]`
- `P3 [:ImageIO], [:Netpbm], [:ImageMagick, LOAD]`

Then the current options are available as fallback, and Netpbm can be used for all formats if ImageIO is not installed.

magister-ludi: I've added these changes.

Thanks @ianshmean and @johnnychen94 for your advice on getting this right.

ianshmean:
I think we're good to merge now, but can you confirm with approval please @johnnychen94

johnnychen94 (approved):

PR #226, opened 2019-04-11
WIP: handle NotInstalledError

johnnychen94:
From previous PR https://github.com/JuliaIO/FileIO.jl/pull/76 this feature looks like supported, but I can't find the commit that deprecates this. And looks like `NotInstalledError` isn't used anymore.

I'll work on test cases if this feature is needed, otherwise, I'll close this PR and clean `error_handling.jl` in next PR.

before:
```julia
julia> using TestImages
[ Info: Recompiling stale cache file /home/math/jc/.julia/compiled/v1.1/TestImages/cMlD2.ji for TestImages [5e47fb64-e119-507b-a336-dd2b206d9990]

julia> testimage("ca")
Error encountered while loading "/home/math/jc/.julia/packages/TestImages/6bT9D/images/cameraman.tif".
Fatal error:
ERROR: ArgumentError: Package ImageMagick not found in current path:
- Run `import Pkg; Pkg.add("ImageMagick")` to install the ImageMagick package.

Stacktrace:
 [1] handle_error(::ArgumentError, ::FileIO.File{FileIO.DataFormat{:TIFF}}) at /home/math/jc/.julia/packages/FileIO/e8FNK/src/error_handling.jl:80
 [2] handle_exceptions(::Array{Any,1}, ::String) at /home/math/jc/.julia/packages/FileIO/e8FNK/src/error_handling.jl:75
 [3] #load#27(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::FileIO.File{FileIO.DataFormat{:TIFF}}) at /home/math/jc/.julia/packages/FileIO/e8FNK/src/loadsave.jl:193
 [4] load(::FileIO.File{FileIO.DataFormat{:TIFF}}) at /home/math/jc/.julia/packages/FileIO/e8FNK/src/loadsave.jl:172
 [5] #load#13(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::String) at /home/math/jc/.julia/packages/FileIO/e8FNK/src/loadsave.jl:118
 [6] load at /home/math/jc/.julia/packages/FileIO/e8FNK/src/loadsave.jl:118 [inlined]
 [7] testimage(::String) at /home/math/jc/.julia/packages/TestImages/6bT9D/src/TestImages.jl:88
 [8] top-level scope at none:0
```

after:

```julia
julia> using TestImages
[...]

PR #287, opened 2020-12-31
Better precompilability

timholy:
These allow more concrete inference and eliminate some weird sources of latency. Very briefly, the first commit compensates for #259 (CC @rofinn, @omus), the last implements the conclusion of discussion in #226 (CC @johnnychen94). All part of a grand scheme to reduce latency (e.g., https://github.com/JuliaGraphics/Gtk.jl/pull/551#issuecomment-752990792).

### DIFF
--- a/src/error_handling.jl
+++ b/src/error_handling.jl
@@ -83,20 +83,5 @@ handle_error(e, q) = throw(e)
 
 function handle_error(e::NotInstalledError, q)
     println("Library \"", e.library, "\" is not installed but is recommended as a library to load format: \"", file_extension(q), "\"")
-    !isinteractive() && rethrow(e) # if we're not in interactive mode just throw
-    while true
-        println("Should we install \"", e.library, "\" for you? (y/n):")
-        input = lowercase(chomp(strip(readline(stdin))))
-        if input == "y"
-            @info(string("Start installing ", e.library, "..."))
-            Pkg.add(string(e.library))
-            return false # don't continue
-        elseif input == "n"
-            @info(string("Not installing ", e.library))
-            return true # User does not install, continue going through errors.
-        else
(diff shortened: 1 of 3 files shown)


rofinn:
The first commit seems fine to me, though I don't fully understand why `filename(q)` isn't inferrable? Would adding a string return type declaration to both the `FileIO.filename` and `FilePathsBase.filename` be a better solution as I can imagine this coming up again?

timholy:
> though I don't fully understand why filename(q) isn't inferrable

It's because `struct File` and `struct Stream` have no type declaration at all for the `filename` field, so as far as inference is concerned it's `::Any`. As far as I understood the point of #259, it was to allow the `filename` field to store something besides `String`...in which case we shouldn't type-assert that an accessor returns a `String`. Or am I misunderstanding?

rofinn:
Oh, I see. It was specifically this change https://github.com/JuliaIO/FileIO.jl/pull/259/files#diff-525588e68b2421901be164965272940effa093de733a3fd36c4b9a4344b8c20cL29 that's causing the problem. The term `filename` confused me as they mean slightly different things between the two packages. Would parameterizing `File` and `Stream` by the type of `filename` in those structs help julia specialize appropriately?

timholy:
Parametrization might be undesirable from a latency perspective: `MyType{T1,T2}` has to be specialized for every *combination* of `T1` and `T2`, and that adds latency. I think the method here is probably better because it doesn't increase the amount of specialization except additively in the number of `filetype(q)` outputs.

timholy:
I don't think this counts as a breaking change: I haven't seen an offer to install a package for years, so I suspect that was dead code. But I'll wait to tag a new release in case someone points out a reason to mark it as breaking.

PR #293, opened 2021-02-25
Add Bundle format with BundlerIO.jl

PaulDebus:
Add support for reading and writing files of the [Bundler format](https://www.cs.cornell.edu/~snavely/bundler/bundler-v0.4-manual.html) used for the results of the structure from motion pipeline


timholy:
Glad they have magic bytes, otherwise that extension `*.out` would make it a beaut' :wink:

PR #290, opened 2021-02-21
Add ImageIO for normal TIFF files

IanButterworth:
Thanks to @tlnagy's https://github.com/tlnagy/TiffImages.jl, which is now added into https://github.com/JuliaIO/ImageIO.jl as of v0.5.0, we now have native _normal_ TIFF IO in julia that we can use in FileIO 🎉

@tlnagy do you have a feel for performance, both IO and load time? Are we ok placing this first for TIFFs?

PR #294, opened 2021-02-26
Try magic bytes func first, if available

IanButterworth:
In https://github.com/JuliaIO/FileIO.jl/pull/290 I noticed that during `load` all magic byte arrays were being tested for, before the magic bytes function.

Given we can identify if there's a magic bytes function earlier, it might make sense to test that before the exhaustive test?

I've only tested this with that ImageIO PR, where load of a small file went from `192.159 μs (414 allocations: 19.67 KiB)` to `125.654 μs (325 allocations: 16.22 KiB)`

### DIFF
--- a/src/query.jl
+++ b/src/query.jl
@@ -80,12 +80,18 @@ function query(filename; checkfile::Bool=true)
         elseif !checkfile && lensym(sym) > 1
             return File{DataFormat{sym[1]}}(filename)
         end
-        if no_magic && !hasfunction(sym)
+        no_function = !hasfunction(sym)
+        if no_magic && no_function
             error("Some formats with extension ", ext, " have no magic bytes; use `File{format\"FMT\"}(filename)` to resolve the ambiguity.")
         end
+        if no_magic && !no_function
+            # try specific function first, if available
+            ret = query(open(filename), abspath(filename), sym)
+            !isnothing(ret) && return file!(ret)
+        end
     end
-    !checkfile && return File{unknown_df}(filename) # (no extension || no magic byte) && no file
-    # Otherwise, check the magic bytes
+    !checkfile && return File{unknown_df}(filename) # (no extension || no magic byte || no function) && no file
+    # Otherwise, check against all magic bytes, then functions
     file!(query(open(filename), abspath(filename)))
 end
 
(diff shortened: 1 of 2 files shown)


src/query.jl
```diff
@@ -80,12 +80,18 @@ function query(filename; checkfile::Bool=true)
...
         end
-        if no_magic && !hasfunction(sym)
+        no_function = !hasfunction(sym)
+        if no_magic && no_function
             error("Some formats with extension ", ext, " have no magic bytes; use `File{format\"FMT\"}(filename)` to resolve the ambiguity.")
         end
+        if no_magic && !no_function
+            # try specific function first, if available
+            ret = query(open(filename), abspath(filename), sym)
+            ret != nothing && return file!(ret)
```
timholy, line 90:
```suggestion
            ret !== nothing && return file!(ret)
```
This ends up calling `===` which is a builtin and hence doesn't use runtime dispatch even in cases like this one where inferrability is nonexistent. For a singleton like `nothing` this is safe.

src/query.jl
```diff
@@ -143,6 +149,28 @@ function query(io::IO, filename = nothing)
     end
     Stream{unknown_df,typeof(io)}(io, filename)
 end
+function query(io::IO, filename::String, sym::Vector{Symbol})
+    magic = Vector{UInt8}()
+    pos = position(io)
+    if seekable(io)
+        for p in filter(x->last(x) in sym, magic_func)
```
timholy, line 156: Instead of creating a new temporary, how about something like
```julia
for (f, fmtsym) in magic_func
    fmtsym in sym || continue
```

timholy (approved):
I like it! It's a very good implementation.

One favor to ask: while you're at this, can you also change the initialization of `magic_func` to
```
const magic_func = Vector{Pair{Function,Symbol}}()
```

src/query.jl
IanButterworth: I repeatedly make this mistake.. thanks

timholy: No worries, and the landscape has already changes since I wrote that: with https://github.com/JuliaLang/julia/pull/38905 merged, starting in Julia 1.7 `isnothing` and `ismissing` could once again become reasonable choices!

timholy: Also it's not really a mistake to use `!=`, it's just a missed opportunity for better inferrability.

PR #296, opened 2021-03-02
Identify TIFF format in IOBuffer

contradict:
IOBuffer does not necessarily have a `:name` property, if it is not present fall back to using only magic bytes.

### DIFF
--- a/test/query.jl
+++ b/test/query.jl
@@ -347,6 +347,17 @@ file_path = Path(file_dir)
     end
 end
 
+@testset "Query from IOBuffer" begin
+    streamformat(::Stream{T, U}) where {T, U} = T
+    for name ∈ ["magic1.tiff", "magic2.tiff"]
+        pth = joinpath(file_dir, "magic2.tiff")
+        open(pth) do io
+            buf=IOBuffer(read(io))
+            @test streamformat(query(buf)) == DataFormat{:TIFF}
+        end
+    end
+end
+
 @testset "Format with function for magic bytes" begin
     add_format(format"FUNCTION_FOR_MAGIC_BYTES", x -> 0x00, ".wav", [:WAV])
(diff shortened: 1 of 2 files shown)


contradict:
Looks like this fails on older Julia versions due to missing `hasproperty`. Is adding a depency on `Compat.jl` an acceptable fix or should I just add an implementation? Is there a better way to check if an `::IO` has a `:name` property?

timholy:
TBH I wouldn't add Compat just for this. How about `:name ∈ propertynames(io)`?

test/query.jl
```diff
@@ -347,6 +347,17 @@ file_path = Path(file_dir)
     end
 end
 
+@testset "Query from IOBuffer" begin
+    streamformat(::Stream{T, U}) where {T, U} = T
+    for name ∈ ["magic1.tiff", "magic2.tiff"]
+        pth = joinpath(file_dir, "magic2.tiff")
+        open(pth) do io
+            buf=IOBuffer(read(io))
+            @test streamformat(query(buf)) == DataFormat{:TIFF}
```
timholy, line 356: How do you know it's TIFF and not OMETIFF? AFAICT only the name distinguishes them.

timholy: Sorry, now I see, you default to TIFF unless it can prove to itself that it's an OMETIFF via `name`. :+1:

contradict: I could also try looking for an XML formatted data block, but adding enough of a TIFF parser to find that didn't seem worthwhile.

timholy:
Thanks!

test/query.jl
tlnagy: Yeah, technically it's only the name that distinguishes the two file formats and the presence of an OME-XML tag inside the TIFF, but that's difficult to probe for in a general package like this one.

PR #298, opened 2021-03-03
Add docs

timholy:
Among other things, this takes the nice table that @SimonDanisch made of the supported formats and runs it as part of every doc build, ensuring we stay up to date.


timholy:
Curiously it didn't make the link, but see preview here: https://juliaio.github.io/FileIO.jl/previews/PR298/

PR #302, opened 2021-03-03
Add support for URIs

### LINKED ISSUE #89: Support URLs
Right now you can't supply a URL for a filename; you have to `download` and then `load`. What do people think about having that be more transparent?

If folks like that: since I'm pretty net-naive, what would be the best thing to search for? Should we require that the string begins with `"http://"` or `"https://"`?

CC @alanedelman

### LINKED ISSUE #219: Support URLs and FilePaths
It'd be nice if the API could be loosened to include [`URI`](https://github.com/JuliaWeb/URIParser.jl/blob/master/src/parser.jl#L13) and [`AbstractPath`](https://github.com/rofinn/FilePathsBase.jl/blob/master/src/FilePathsBase.jl#L84) types.

Related issues: #89, https://github.com/rofinn/FilePaths.jl/issues/34, https://github.com/rofinn/FilePaths.jl/pull/22#issuecomment-370620375

timholy:
User must load HTTP manually for this to be activated.
(It exploits Requires.)

Closes #219
Closes #89

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "1.6.0"
 
 [deps]
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
@@ -12,8 +13,9 @@ julia = "0.7, 1"
 [extras]
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
 FilePathsBase = "48062228-2e41-5def-b9a4-89aafe57970f"
+HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["ColorTypes", "FilePathsBase", "Random", "Test"]
+test = ["ColorTypes", "FilePathsBase", "HTTP", "Random", "Test"]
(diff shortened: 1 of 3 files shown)


johnnychen94:
FWIW, ping @oxinabox since ReferenceTests requires FileIO.

timholy:
Out of curiosity, what don't you like? I used to dislike it too until we got the latency down...but oh, it's crept back up again. [works for a while] There, that's better: https://github.com/JuliaPackaging/Requires.jl/pull/101

SimonDanisch:
Isn't Download.jl a std lib in 1.6, so basically free to depend on going forward - and also supposed to be the stable standard to download files?

timholy:
Yes, but I'm not sure I understand. Are you saying we should move Requires to Download?

SimonDanisch:
No, use Download instead of HTTP and just depend on it without Requires... Although I just see, that we would also need to depend on URIParsers.jl in that case (but it's at least also a small, dependency free package).

timholy:
Oh, interesting, I'd not noticed that. That seems like a good idea if we stop supporting anything younger than 1.6.

Indeed, one fortunate thing is that the most recent release of HTTP factored out their URI support into URIs.jl, and URIParsers.jl is deprecated. If we can rely on a minimum release that depends on URIs, we can make a seamless transition from HTTP to URIs.

If you have a dependency which might not be used:
```julia
tim@diva:~/src/julia-master/base$ juliam --startup-file=no -q
julia> @time using Example    # precompile the package-loading machinery
  0.047376 seconds (103.68 k allocations: 6.508 MiB, 89.74% compilation time)

julia> @time using URIs
  0.065603 seconds (102.69 k allocations: 6.830 MiB)

julia> 
tim@diva:~/src/julia-master/base$ juliam --startup-file=no -q
julia> @time using Example     # fresh session, do it again
  0.046348 seconds (103.68 k allocations: 6.508 MiB, 89.96% compilation time)

julia> @time using Requires
  0.052093 seconds (21.19 k allocations: 1.871 MiB)
```
but together with Requires' `__init__` it goes to ~150ms (see demo in https://github.com/JuliaPackaging/Requires.jl/pull/101).

PR #309, opened 2021-03-08
Add compat note re File/Stream

timholy:
Suggested in
https://github.com/JuliaStats/RDatasets.jl/pull/119#discussion_r589220963


timholy:
CC @devmotion

devmotion:
Thanks, looks good! Hopefully it helps others to avoid the mistake I made :slightly_smiling_face:

timholy:
It's not a mistake if it's not documented...

Issue #310, opened 2021-03-10
v1.6.0 broke VegaLite.jl and Vega.jl


davidanthoff:
The major pattern for saving plots in VegaLite.jl is this:
```
using VegaDatasets, VegaLite, FileIO

dataset("cars") |>
  @vlplot(:point, x=:Horsepower,y=:Miles_per_Gallon,color=:Origin,width=400,height=400) |>
  save("foo.png")
```

That stopped working with FileIO.jl 1.6.0. I've tried to narrow down what is happening, and I think the problem is that [here](https://github.com/JuliaIO/FileIO.jl/blob/b779539b505ba5aea6634c5cb78b2d2cc6e49ac8/src/loadsave.jl#L148) we end up with a list that does not include the `MimeWriter` (which is the writer that should be used for that scenario).

I actually managed to use the debugger to figure that out :) But beyond that I ran into all of these functions that were generated with `@eval` (like [here](https://github.com/JuliaIO/FileIO.jl/blob/b779539b505ba5aea6634c5cb78b2d2cc6e49ac8/src/loadsave.jl#L8)) and then I was no longer able to follow what is going on. So this is an entirely unrelated side comment, but I've really come around to think that using these kind of macro code generation tricks in situations like that is not a great idea :) It might save the original author of the code a bit of time, but for folks like me, that just want to fix some (presumably pretty simple) bug, this kind of stuff makes it really sooo much more difficult to get into a code-base.

Would be great if we could try to fix this with a priority, because this has really broken VegaLite.jl pretty thoroughly :)

CC @timholy 

EDIT: Had used `@save` originally, that should have been `save`.

IanButterworth:
In 1.7 I can't find the function `@save`. Is that the error you're getting? (if not, do you get an error message  / stack trace?)

```
(jl_CL5U7e) pkg> add VegaDatasets, VegaLite, FileIO
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
   Resolving package versions...
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_CL5U7e/Project.toml`
  [5789e2e9] + FileIO v1.6.2
  [0ae4a718] + VegaDatasets v2.1.1
  [112f6efa] + VegaLite v2.3.1
...

julia> using VegaDatasets, VegaLite, FileIO

julia> dataset("cars") |>
[...]

PR #311, opened 2021-03-11
Fix typo in save method

IanButterworth:
Edit: It's just a typo

___

I think exposure of the MimeWriter save functions may have inadvertently dropped off in the refactor, with the change from

https://github.com/JuliaIO/FileIO.jl/blob/71bdffe1f3933aa51e7d7c42f50b90589e52d7f4/src/loadsave.jl#L143-L144

to 

https://github.com/JuliaIO/FileIO.jl/blob/b779539b505ba5aea6634c5cb78b2d2cc6e49ac8/src/loadsave.jl#L145-L150

At least, seems to fix https://github.com/JuliaIO/FileIO.jl/issues/310 for me.

I very much defer to @timholy's judgement here though about whether this is the right way to fix it

cc. @davidanthoff

### DIFF
--- a/src/loadsave.jl
+++ b/src/loadsave.jl
@@ -145,7 +145,7 @@ end
 # return a save function, so you can do `thing_to_save |> save("filename.ext")`
 function save(file; options...)
     sym = querysym(file; checkfile=false)
-    libraries = applicable_loaders(sym)
+    libraries = applicable_savers(sym)
     return data -> action(:save, libraries, sym, file, data; options...)
 end
 


davidanthoff:
Yes, this does fix it for me as well! Thanks a lot!!

@timholy any chance we could merge this quickly and tag a new release?

timholy:
I'm confused about this. We do have https://github.com/JuliaIO/FileIO.jl/blob/b779539b505ba5aea6634c5cb78b2d2cc6e49ac8/src/loadsave.jl#L146-L150

When I put some debugging code in ImageMagick, the passed in image is a `VegaLite.VLSpec`. Why isn't is an image?

src/loadsave.jl
```diff
@@ -218,5 +218,15 @@ function action(call::Symbol, libraries::Vector{ActionSource}, @nospecialize(fil
             push!(failures, (e, library, catch_backtrace()))
         end
     end
+    if call === :save # Try MimeWriter methods
+        try
+            return MimeWriter.save(file, args...; options...)
```
timholy, line 223: You're overriding the registry to force MimeWriter. It's also listed as a valid format for PNGs: https://github.com/JuliaIO/FileIO.jl/blob/b779539b505ba5aea6634c5cb78b2d2cc6e49ac8/src/registry.jl#L146

So why doesn't that work?

timholy:
Sorry not to go farther in debugging this but it's been a crazy several days and I need to go celebrate my wife's birthday...but I am skeptical about merging this pronto. Really sorry for the breakage.

IanButterworth:
Forget everything I did. Looks like it was just a typo.. `applicable_loaders` -> `applicable_savers`.
Seems to work for me..

Sorry for the noise

src/loadsave.jl
IanButterworth: ignore everything I did please 😂

davidanthoff:
Ha, that is an easy fix :) And amazing how one can _not_ spot things like that. I must have stared at that line for a significant time when I was debugging that, but of course it did _not_ at all jump at me what was going on there :)

timholy:
:disappointed_relieved: OMG. I'll merge this and add a test in a separate PR.

timholy:
Thanks for fixing this!

PR #312, opened 2021-03-12
Add a functional-form MimeWriter test

timholy:
This would have prevented #310 (which was fixed in #311).

CC @IanButterworth, @davidanthoff


davidanthoff:
Great!

PR #313, opened 2021-03-12
fix detectwav

jkrumbiegel:


timholy:
Oops, my fault. Thanks! Test added in #314

Issue #225, opened 2019-04-08
Support for gz-compressed RData files


alyst:
In JuliaData/RData.jl#57 we have ran into problems with gzip-compressed R Data file with non-registered extension (`.Rda`). Since none of FileIO-registered extensions were matched, `load()` was trying to detect the format by looking into magic bytes. Naturally, the file was detected as gzip archive.

Would it be possible to fix the way RData files are registered in FileIO, so that gzipped files with non-standard extension are correctly recognized as RData? Is there a generic way to specify that the given file format could exist in compressed and uncompressed forms? (It would be useful not only for RData, but e.g. for compressed XML, SVG)

SimonDanisch:
previous discussion: https://github.com/JuliaIO/FileIO.jl/issues/32

PR #318, opened 2021-03-17
Fix detect_compressed()

alyst:
The PR fixes a typo in BZIP2 magic bytes that is responsible for JuliaData/RData.jl#83 and adds unit tests for all supported compression formats.

I've also refactored the compression detection: now there is `detect_compressor()` function that returns the compression format code or *nothing* if no supported format detected.
I would be using it from *RData.jl* instead of the compression detection code that is currently there to avoid code duplication.
Also, it could help in future when implementing the format detection of decompressed streams.
For example, the current version of *FileIO* checks for *RData*-specific magic bytes if the file is not compressed, but for compressed *RData* files it just checks the compression format but skips magic bytes check of the uncompressed stream (see also #225).

Once merged, it would be nice to tag a new version of *FileIO*, because the fix of JuliaData/RData.jl#83 depends on it.


alyst:
@timholy could you please take a look?

johnnychen94 (approved):
Curious to ask, will it be better to use `:GZIP`, `:BZIP2`, `:XZ`, `:LZ4` and perhaps `:plain` (or `:uncompressed`) here?

johnnychen94:
> Once merged, it would be nice to tag a new version of FileIO, because the fix of JuliaData/RData.jl#83 depends on it.

FileIO v1.6.5 is out.

alyst:
> Curious to ask, will it be better to use `:GZIP`, `:BZIP2`, `:XZ`, `:LZ4` and perhaps `:plain` (or `:uncompressed`) here?

Thanks for merging!
I think returning `:uncompressed` would be misleading as we only check a limited subset of compressors.
I don't have a particular preference to symbols or strings, I've just kept the current choice.

PR #319, opened 2021-04-04
add AVI MP4 OGG and Matroska video container support via VideoIO

IanButterworth:
As of `VideoIO` v0.9.1 there are simple `VideoIO.save` and `VideoIO.load` functions, which are extended to the `fileio_save` `fileio_load` names here
https://github.com/JuliaIO/VideoIO.jl/blob/77091893d21aae80cf7bca2c8550f0b55f029b36/src/VideoIO.jl#L111-L113

This adds basic functionality to load the entire video as a framestack, or write an entire framestack to file in one go.

FFMPEG can support a lot of file/container/codec types, but I thought I'd start with the common ones.

I tried to figure out how to add .mov but the magic bytes are a bit confusing. Happy to add them if someone figures out a decent heuristic.

```julia
julia> @time vid = FileIO.load("/Users/ian/Documents/GitHub/VideoIO.jl/videos/ladybird.mp4");
  4.754831 seconds (3.33 k allocations: 2.300 GiB, 6.35% gc time)

julia> typeof(vid)
Vector{PermutedDimsArray{RGB{N0f8}, 2, (2, 1), (2, 1), Matrix{RGB{N0f8}}}} (alias for Array{PermutedDimsArray{ColorTypes.RGB{FixedPointNumbers.Normed{UInt8, 8}}, 2, (2, 1), (2, 1), Array{ColorTypes.RGB{FixedPointNumbers.Normed{UInt8, 8}}, 2}}, 1})

julia> size(vid)
(397,)

julia> size(vid[1])
(1080, 1920)

julia> @time FileIO.save("/Users/ian/Documents/GitHub/VideoIO.jl/videos/ladybird2.mp4", vid);
 10.289047 seconds (1.71 k allocations: 30.203 KiB)

julia> @time FileIO.save("/Users/ian/Documents/GitHub/VideoIO.jl/videos/ladybird2.mp4", vid, encoder_options = (color_range=2, crf=0, preset="ultrafast"));
  3.476963 seconds (1.73 k allocations: 30.797 KiB)
```

Note that this puts VideoIO ~before~ AFTER the existing ImageMagick AVI support. ~We may want to be more conservative and leave ImageMagick in first place?~

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -1,7 +1,7 @@
 function _precompile_()
     ccall(:jl_generating_output, Cint, ()) == 1 || return nothing
     for f in (detect_rdata, detect_rdata_single, detectwav, detect_bedgraph,
-              detecttiff, detect_noometiff, detect_ometiff, detect_avi,
+              detecttiff, detect_noometiff, detect_ometiff, detectavi,
               detecthdf5, detect_stlascii, detect_stlbinary, detect_gadget2)
         @assert precompile(f, (IOStream,))
     end
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -1,7 +1,7 @@
 function _precompile_()
     ccall(:jl_generating_output, Cint, ()) == 1 || return nothing
     for f in (detect_rdata, detect_rdata_single, detectwav, detect_bedgraph,
-              detecttiff, detect_noometiff, detect_ometiff, detectavi,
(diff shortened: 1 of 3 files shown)


IanButterworth:
I'm going to merge given this doesn't break anything and just adds functionality.

I had renamed `detectavi` to `detect_avi` to conform to standard practice, and put VideoIO ahead of the existing ImageMagick support for AVI,  but I think it's better to be conservative, so I reverted those changes.

PR #321, opened 2021-04-04
Do not use isfile for URLs in checkpath_load

### LINKED ISSUE #320: The 1.6 update seems to have broken load from URLs for CSVFiles
I think this must be another breaking change that was introduced in 1.6: CSVFiles.jl forever has supported loading of files from URLs, i.e. something like `load("https://raw.githubusercontent.com/queryverse/CSVFiles.jl/v0.2.0/test/data.csv")` used to work. With the new release of 1.6 of FileIO this now throws an error `ArgumentError: No file exists at given path: https://raw.githubusercontent.com/queryverse/CSVFiles.jl/v0.2.0/test/data.csv`, I think at [this](https://github.com/JuliaIO/FileIO.jl/blob/d203090ea7d07985232c8b5130d50178f929b84f/src/loadsave.jl#L167) location.

That check should probably only happen if the `file` argument references a local path?

CC @timholy because blame says you changed this :)

davidanthoff:
Fixes #320. Not sure this is the best solution, but it would still be great if we could merge this quickly and tag a new release, so that the functionality in CSVFiles.jl is no longer broken.

### DIFF
--- a/src/loadsave.jl
+++ b/src/loadsave.jl
@@ -164,7 +164,7 @@ end
 function checkpath_load(file)
     file === nothing && return nothing   # likely stream io
     isa(file, IO) && return nothing
-    !isfile(file) && throw(ArgumentError("No file exists at given path: $file"))
+    startswith(file, "http://") || startswith(file, "https://") || !isfile(file) && throw(ArgumentError("No file exists at given path: $file"))
     return nothing
 end
 function checkpath_save(file)


davidanthoff:
@timholy Hm, I don't fully understand what is going on here. The test errors seem to indicate that this package is now somehow integrated with FilePaths.jl? I don't really know how to continue here, but I do know that 1.6 broke things on this aspect of things :) Any ideas how we could resolve this?

src/loadsave.jl
```diff
@@ -164,7 +164,7 @@ end
 function checkpath_load(file)
     file === nothing && return nothing   # likely stream io
     isa(file, IO) && return nothing
-    !isfile(file) && throw(ArgumentError("No file exists at given path: $file"))
+    startswith(file, "http://") || startswith(file, "https://") || !isfile(file) && throw(ArgumentError("No file exists at given path: $file"))
```
johnnychen94, line 167: How about using a new helper function?

```julia
is_url(path::AbstractString) = startswith(file, "http://") || startswith(file, "https://")
is_url(path) = false
```

johnnychen94, line 167: The logic seems a little bit hard to understand. If I understand it correctly, is the following an equivalent version?

```suggestion
    startswith(file, "http://") || startswith(file, "https://") || isfile(file) || throw(ArgumentError("No file exists at given path: $file"))
```

johnnychen94:
Tests are needed to avoid #320

davidanthoff:
@timholy Bump :) The 1.6 update here broke functionality in CSVFiles.jl that had been working for years, would be good if we could somehow fix that, but I'm stuck, I don't understand what changes were introduced in FileIO.jl that caused this in the first place...

PR #326, opened 2021-04-28
Fix collision in case insensitive file systems

### LINKED ISSUE #325: FIle collision on case sensitive file systems after recent release.
https://github.com/JuliaIO/FileIO.jl/pull/323 added some test files that have collisions one case insensitive file systems.

In particular, the two files `'test/files/doxy.MID'` and `'test/files/doxy.mid'` only differ in the casing.

This causes the following problem:

```
base ❯ git clone https://github.com/JuliaIO/FileIO.jlCloning into 'FileIO.jl'...
...
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'test/files/doxy.MID'
  'test/files/doxy.mid'
```

This also confuses some Pkg tree hash routine so whenever someone installs FileIO they now get:

```
┌ Warning: tarball content does not match git-tree-sha1, for package downloaded from https://pkg.julialang.org/package/5789e2e9-d7fb-5bc7-8068-2c6fae9b9549/b647ed22f176a4f12eabe2afd90d19a95af2e8ca, expected b647ed22f176a4f12eabe2afd90d19a95af2e8ca got 8dad1c042a8fadf62c5ea801a93abf54d00eeecb
```

The solution is to rename one of the files and make a new release.

cc @johnnychen94, @VasanthManiVasi

VasanthManiVasi:
Renamed doxy.MID to doxy2.MID since it caused problems in case insensitive file systems as mentioned in #325. Closes #325.


KristofferC (approved):
Would be good with a new release after this

PR #327, opened 2021-04-30
relax `isfile` check to allow URL input in `checkpath_load`

### LINKED ISSUE #320: The 1.6 update seems to have broken load from URLs for CSVFiles

johnnychen94:
closes #320 closes #321

cc: @davidanthoff

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,6 +12,7 @@ Requires = "1"
 julia = "0.7, 1"
 
 [extras]
+CSVFiles = "5d742f6a-9f54-50ce-8119-2520741973ca"
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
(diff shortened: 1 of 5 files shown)


davidanthoff:
Fantastic, much appreciated!!

PR #330, opened 2021-05-06
Extend support for ISO/IEC base media format

contradict:
MP4 files are often encoded using this format, the first 4 bytes are a
box length and should not be part of the file identity magic.

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -188,7 +188,23 @@ function detectavi(io)
     submagic == b"AVI "
 end
 add_format(format"AVI", detectavi, ".avi", [idImageMagick], [idVideoIO])
-add_format(format"MP4", UInt8[0x00,0x00,0x00,0x18,0x66,0x74,0x79,0x70], ".mp4", [idVideoIO])
+
+""" detectisom(io)
+
+Detect ISO/IEC 14496-12 ISO/IEC base media format files. These files start with
+a 32-bit big-endian length, and then the string 'ftyp' which is followed by
+details of the container and codec. Finding 'ftyp' is enough to know to dispatch
+to VideoIO.
+"""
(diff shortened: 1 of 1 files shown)


src/registry.jl
```diff
@@ -188,7 +188,23 @@ function detectavi(io)
...
 end
 add_format(format"AVI", detectavi, ".avi", [idImageMagick], [idVideoIO])
-add_format(format"MP4", UInt8[0x00,0x00,0x00,0x18,0x66,0x74,0x79,0x70], ".mp4", [idVideoIO])
+
+""" detectisom(io)
+
+Detect ISO/IEC 14496-12 ISO/IEC base media format files. These files start with
+a 32-bit big-endian length, and then the string 'ftyp' which is followed by
+details of the container and codec. Finding 'ftyp' is enough to know to dispatch
+to VideoIO.
```
johnnychen94, line 197: > Finding 'ftyp' is enough to know to dispatch to VIdeoIO

I'm not an expert on video formats. Are there any references for this that can help me/others to clarify this?

johnnychen94 (approved):
No objections but would be better to have @IanButterworth double-check it.

src/registry.jl
contradict: I'm not an expert either, I used these three references to come to this conclusion:
Wikipedia has a readable introduction: https://en.wikipedia.org/wiki/ISO/IEC_base_media_file_format#File_type_box

There is a publicly available version of the spec here: https://standards.iso.org/ittf/PubliclyAvailableStandards/c068960_ISO_IEC_14496-12_2015.zip

As a check, I compared to the C++ code here: https://github.com/DigiDNA/ISOBMFF

PR #323, opened 2021-04-22
Add MIDI format

VasanthManiVasi:
Hi! I've added the support for MIDI formats. The I/O implementation is here at JuliaMusic/MIDI.jl/pull/140. Should this PR be merged before the MIDI.jl PR gets merged? Would love your help on which one to do first!

PR #333, opened 2021-05-13
[registry] AVS .fld data files

JeffFessler:
This PR adds support for AVS .fld data files, using the new library
https://github.com/JeffFessler/AVSfldIO.jl
(File format etc. is documented there with links to many software frameworks that support it.)
The library is in the middle of the 3-day waiting period for the registry.
I have 100% code coverage in the library but I'm unsure if I need to add any tests to FileIO.jl.
Update: I added magic bytes.


JeffFessler:
BTW, I'd be happy to have AVSfldIO.jl housed under JuliaIO.  Please advise.

johnnychen94:
> Is 1.8.3 the right version or is it 1.9.0 now?

Yes, it should be v1.9.0 since it's a new feature. But given that there're also other new formats registered as pending PRs, it's better to modify the version number outside of these PRs as an independent commit. Can you revert the version change in Project.toml?

> There are no magic bytes for this file format.
> But I could write a simple detector if it is required for merging.

Yes, I think a header check is needed as a detector. See comments in : https://github.com/JuliaIO/FileIO.jl/pull/222#issuecomment-563988521

> BTW, I'd be happy to have AVSfldIO.jl housed under JuliaIO. Please advise.

ping @ianbutterworth, do we need to invite @JeffFessler as a member first?

JeffFessler:
@johnnychen94 I realized that actually this format does have magic bytes, namely "# AVS" so I have updated the PR to include that.  So now I think that no further header check is needed.
I also reverted the version as requested.
Please advise what type of test to add and where to put it.  I find it hard to follow the testing code in `test/` here.  Probably `loadsave.jl` is the file to edit, but it is 400 lines with no obvious place for adding new formats.  (Perhaps a separate file of test code for each new type would be more clear?)

johnnychen94:
> Please advise what type of test to add and where to put it. I find it hard to follow the testing code in test/ here. Probably loadsave.jl is the file to edit, but it is 400 lines with no obvious place for adding new formats. (Perhaps a separate file of test code for each new type would be more clear?)

A good question... The current case from my understanding is that some add tests and some don't. I prefer to have tests because it makes things more reliable.

People add `query` tests here in `test/query.jl` so as to make sure `FileIO` correctly calls one of the available backends. You can take a look at #323 as an example.

test/files/avs-ascii.fld
```diff
@@ -0,0 +1,9 @@
+# AVS field file (ascii variant)
```
johnnychen94, line 1: Just a question. Does all AVS field file has this `# AVS`?

johnnychen94 (approved):

test/files/avs-ascii.fld
JeffFessler: Every one that I have seen.  This format was never blessed by any standards organization and I have been unable to find online documentation of it.  (I learned the format from a paper manual before the WWW existed!)  So I think we will not find any 100% definitive answer to the question.
I already added a comment about this assumption in the long README here:
https://github.com/JeffFessler/AVSfldIO.jl
That README is probably more complete documentation than you will find anywhere on the web!

It is trivial to add that header if someone (1) has a file without it and (2) wants to read it with Julia.
If at some later point there is someone who has a big pile of the data format without that header then I am happy to write a fancier format detector function if needed.

I have added the query test in the most recent commit.
I also changed the format name from "FLD" to "AVSfld" to be more specific.

Issue #336, opened 2021-05-26
FileIO is not threadsafe


c42f:
Debugging some concurrency issues, I ended up inside the source of `FileIO` and saw that it maintains global state in `sym2loader` and `sym2safer` (and maybe other places) which is not protected by locks.

Presumably this can cause problems when these dictionaries are mutated during `load`, etc.

I don't have time to make an MWE right now, but I thought I'd leave a note about this nevertheless.

Here's a sample stacktrace from a loop which essentially does the following for a bunch of JPEG files:

```julia
@sync for f in files
    @spawn open(f) do io
        load(io)
    end
end
```

```
ERROR: TaskFailedException
    nested task error: concurrency violation detected
    Stacktrace:
      [1] error(s::String)
        @ Base ./error.jl:33
      [2] concurrency_violation()
        @ Base ./condition.jl:8
      [3] assert_havelock
        @ ./condition.jl:25 [inlined]
      [4] assert_havelock
        @ ./condition.jl:48 [inlined]
      [5] assert_havelock
        @ ./condition.jl:72 [inlined]
      [6] wait(c::Condition)
        @ Base ./condition.jl:102
      [7] _require(pkg::Base.PkgId)
        @ Base ./loading.jl:978
      [8] require(uuidkey::Base.PkgId)
        @ Base ./loading.jl:914
      [9] action(::Symbol, ::Vector{Union{Base.PkgId, Module}}, ::Formatted; options::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
        @ FileIO ~/data/.julia/packages/FileIO/3jBq2/src/loadsave.jl:202
     [10] action
        @ ~/data/.julia/packages/FileIO/3jBq2/src/loadsave.jl:196 [inlined]
     [11] #load#15
        @ ~/data/.julia/packages/FileIO/3jBq2/src/loadsave.jl:120 [inlined]
     [12] load
        @ ~/data/.julia/packages/FileIO/3jBq2/src/loadsave.jl:117 [inlined]
     [13] #7
        @ /mnt/data/code/matt_test.jl:13 [inlined]
     [14] open(f::var"#7#10", args::String; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
        @ Base ./io.jl:330
     [15] open(f::Function, args::String)
        @ Base ./io.jl:328
[...]

PR #339, opened 2021-06-02
move package loading behind a lock

KristofferC:
Ref https://github.com/JuliaIO/FileIO.jl/issues/336, not sure if it completely fixes the issue.

### DIFF
--- a/src/loadsave.jl
+++ b/src/loadsave.jl
@@ -192,6 +192,7 @@ const fileiofuncs = Dict{Symbol,Function}(:load => load,
                                           :save => save,
                                           :savestreaming => savestreaming)
 
+const require_lock = ReentrantLock()
 function action(call::Symbol, libraries::Vector{ActionSource}, @nospecialize(file::Formatted), args...; options...)
     issave = call ∈ (:save, :savestreaming)
     failures = Tuple{Any,ActionSource,Vector}[]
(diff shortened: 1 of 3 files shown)


johnnychen94 (approved):
> Perhaps this indicates that Base.require itself is not threadsafe, because it appears to use plain Base.Condition, rather than Threads.Condition.

Just curious, could this be fixed in upstream Julia?

KristofferC:
Maybe, but I don't think that lock is really for concurrency safety but to limit recursive `require` calls. But I am not sure.

johnnychen94:
I'll set up the CI and check if it works; I have good reason to believe this is the right fix but we still need to appropriately test it.

johnnychen94:
I've locally verified that this test 1c9db8f captures the #336, so I'll merge once CI passes.

src/loadsave.jl
```diff
@@ -192,14 +192,15 @@ const fileiofuncs = Dict{Symbol,Function}(:load => load,
                                           :save => save,
                                           :savestreaming => savestreaming)
 
+const require_lock = ReentrantLock()
```
johnnychen94, line 195: It looks like for Julia <= v1.3, `Threads.SpinLock()` is required here.

johnnychen94:
Some strange behavior here when testing CSVFiles: `SystemError: dup: Bad file descriptor`. using `SpinLock` doesn't workaround it. Any ideas?

cc: @davidanthoff

johnnychen94:
Given that multi-threads are not very well supported before Julia 1.3, I decided to only test this with Julia >= 1.3.

I've locally tested it for all minor versions from 1.3-1.6 on my macOS and Ubuntu. For Julia 1.1 it can be fixed by using `SpinLock`, but for Julia 1.0 and 1.2, it is still broken anyway.

This is only a partial fix to #336 for Julia >= 1.3.

davidanthoff:
@johnnychen94 sorry, only catching up right now with notifications. Was the problem with CSVFiles.jl resolved before this was merged? Or is there still something I should look into?

johnnychen94:
@davidanthoff It was then identified an unrelated issue. Sorry for the spam!

PR #340, opened 2021-06-04
[registry] add MuData and AnnData formats

ilia-kats:
Depends on [Muon.jl](https://github.com/gtca/Muon.jl). The formats are HDF5-based and only MuData always has a custom header (AnnData can also be just a plain HDF5 file), therefore the awkwardness with ordering the `add_format` calls.


johnnychen94:
This is a hard call to me. Correct me if I'm wrong. I don't think we should assume the registry is ordered; the priority defined here is expected to be an undefined behavior to me.

I'll see how others think of this usage. (We don't have many active people here so it might take times).

ilia-kats:
The registry is implemented in an ordered fashion, at least. Besides, there is precedent: JLD2 is also using HDF5 files with a custom header, the only reason why JLD2 files always use the JLD2 functions instead of the generic HDF5 ones is because JLD2 is added at the top of the registry file and uses the ordered registry assumption.

src/registry.jl
```diff
@@ -381,8 +381,23 @@ function detecthdf5(io)
...
     false
 end
+function detect_mudata(io)
+    seekstart(io)
+    if String(read(io, 6)) != "MuData"
+        return false
+    end
+    seekstart(io)
+    return detecthdf5(io)
+end
```
johnnychen94, line 391: Make it all bytes comparison to reduce overhead.

```suggestion
const MUON_MAGIC = UInt8['M', 'u', 'D', 'a', 't', 'a']
function detect_mudata(io)
    seekstart(io)
    match(io, MUON_MAGIC) || return false
    seekstart(io)
    return detecthdf5(io)
end
```

johnnychen94 (commented):
Would you mind adding some query check tests to `test/query.jl` to make sure `h5mu`, `h5ad`, and `hdf5` query results are expected?

With proper tests, I think it's okay to get it in.

@ilia-kats I'm preparing for a new release #344 so it would be great if you can make it recently, or let me know if you're not available.

johnnychen94 (approved):

PR #348, opened 2021-09-08
Limit output when printing errors

timholy:
For some errors, the entire data set being saved can be printed to
the terminal when an error is thrown. This  reduces the output.


timholy:
Failure fixed in #349

PR #349, opened 2021-09-08
Workaround windows failure

timholy:
For reasons not understood, Windows fails to download the test files
under Julia 1.0. It would be great if someone debugged this, but in the
meantime it seems best to work around the failure.


timholy:
Example: https://github.com/JuliaIO/FileIO.jl/pull/348/checks?check_run_id=3542927418

timholy:
OSX stuck. Merge anyway.

Issue #350, opened 2021-10-06
AstroIO required when it shouldn't be


junglegobs:
I load a lot (~10,000) JLD2 files using `collect_results` from [DrWatson.jl](https://juliadynamics.github.io/DrWatson.jl/dev/). For a couple of files it's fine, but for the ~10,000 that I just mentioned I get the error that `AstrioIO` is required. This strikes me as very strange, since it's a rather obscure package which I definitely don't need to load a JLD2 file.

Stacktrace:

```julia
Error encountered while load FileIO.File{FileIO.DataFormat{:Gadget2}, String}("/home/u0128861/Desktop/storage-and-LOLE/results/monte_carlo/default_nP=1/mc_year_1016.jld2").

Fatal error:
ERROR: LoadError: ArgumentError: Package AstroIO [c85a633c-0c3f-44a2-bffe-7f9d0681b3e7] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.

Stacktrace:
  [1] _require(pkg::Base.PkgId)
    @ Base ./loading.jl:1012
  [2] require(uuidkey::Base.PkgId)
    @ Base ./loading.jl:936
  [3] #34
    @ ~/.julia/packages/FileIO/DNKwN/src/loadsave.jl:203 [inlined]
  [4] lock(f::FileIO.var"#34#35"{Base.PkgId}, l::ReentrantLock)
    @ Base ./lock.jl:187
  [5] action(::Symbol, ::Vector{Union{Base.PkgId, Module}}, ::FileIO.Formatted; options::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ FileIO ~/.julia/packages/FileIO/DNKwN/src/loadsave.jl:203
  [6] action
    @ ~/.julia/packages/FileIO/DNKwN/src/loadsave.jl:197 [inlined]
  [7] action(::Symbol, ::Vector{Union{Base.PkgId, Module}}, ::Symbol, ::String; options::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ FileIO ~/.julia/packages/FileIO/DNKwN/src/loadsave.jl:185
  [8] action
    @ ~/.julia/packages/FileIO/DNKwN/src/loadsave.jl:185 [inlined]
[...]

PR #351, opened 2021-10-26
More concrete Gadget2 format detector

islent:
Probably fix https://github.com/JuliaIO/FileIO.jl/issues/350


johnnychen94:
Let's test this in the wild.

PR #328, opened 2021-05-03
[registry] Add OpenEXR via ImageIO

twadleigh:
Addresses https://github.com/twadleigh/OpenEXR.jl/issues/2

Depends on https://github.com/JuliaIO/ImageIO.jl/pull/30

PR #354, opened 2021-12-27
add QOI.jl to read and save the qoi image format

KristofferC:
Requires QOI.jl to be registered and https://github.com/KristofferC/QOI.jl/pull/2 to be merged

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -489,3 +489,5 @@ add_format(format"FCS", "FCS", [".fcs"], [:FCSFiles => UUID("d76558cf-badf-52d4-
 add_format(format"HTML", (), [".html", ".htm"], [MimeWriter, SAVE])
 
 add_format(format"MIDI", "MThd", [".mid", ".midi", ".MID"], [:MIDI => UUID("f57c4921-e30c-5f49-b073-3f2f2ada663e")])
+
+add_format(format"QOI", "qoif", ".qoi", [:QOI => UUID("4b34888f-f399-49d4-9bb3-47ed5cae4e65")])
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -490,4 +490,4 @@ add_format(format"HTML", (), [".html", ".htm"], [MimeWriter, SAVE])
 
 add_format(format"MIDI", "MThd", [".mid", ".midi", ".MID"], [:MIDI => UUID("f57c4921-e30c-5f49-b073-3f2f2ada663e")])
 
-add_format(format"QOI", "qoif", ".qoi", [:QOI => UUID("4b34888f-f399-49d4-9bb3-47ed5cae4e65")])
+add_format(format"QOI", "qoif", ".qoi", [:QOI => UUID("4b34888f-f399-49d4-9bb3-47ed5cae4e65")], [idImageIO])


johnnychen94:
Love this! I was planning to give it a try during this winter vacation; maybe we can also include it in ImageIO and ship it to all users of Images?

KristofferC:
Sounds good to me!

KristofferC:
@johnnychen94 This should be good to go.

KristofferC:
Or @timholy.

src/registry.jl
```diff
@@ -489,3 +489,5 @@ add_format(format"FCS", "FCS", [".fcs"], [:FCSFiles => UUID("d76558cf-badf-52d4-
 add_format(format"HTML", (), [".html", ".htm"], [MimeWriter, SAVE])
 
 add_format(format"MIDI", "MThd", [".mid", ".midi", ".MID"], [:MIDI => UUID("f57c4921-e30c-5f49-b073-3f2f2ada663e")])
+
+add_format(format"QOI", "qoif", ".qoi", [:QOI => UUID("4b34888f-f399-49d4-9bb3-47ed5cae4e65")])
```
johnnychen94, line 493: We're trying to make ImageIO a replacement for ImageMagick and make ImageIO the only interface for normal users, would you mind registering this package to ImageIO instead?

An OpenEXR example can be found in https://github.com/JuliaIO/ImageIO.jl/pull/30 and https://github.com/JuliaIO/FileIO.jl/pull/328.

johnnychen94 (commented):
Because this format doesn't have magic numbers (does it?), I think we need to have some tests here to ensure we don't break it in the future.

---

Curious to ask, how stable do you think the current QOI specification (not QOI.jl implementation) is stable enough? This format is quite promising but it seems to me that it's still experimental.

KristofferC:
It does have magic numbers (which this PR) adds. The format is done and new changes are not accepted. So the qoi format will stay like this.

src/registry.jl
KristofferC: https://github.com/JuliaIO/ImageIO.jl/pull/42, so does this mean it should not be registered in FileIO? I don't immidiately see the benefit of ImageIO over just using FileIO. Should I still have https://github.com/KristofferC/QOI.jl/blob/577645ab3e5930231f4043543d1c622824ceca48/src/QOI.jl#L364-L365 in the package? Should this PR be merged in addition to the ImageIO one, or only ImageIO?

johnnychen94: > JuliaIO/ImageIO.jl#42, so does this mean it should not be registered in FileIO? I don't immediately see the benefit of ImageIO over just using FileIO.

That's a good question. I'm not sure if my answer satisfies you: the primary goal is to provide an image-related IO toolbox as an alternative to ImageMagick so that when new users come up, we can save the efforts and tell them to `]add ImageIO` instead of `]add PNGFiles QOI ...`.
The latter can sometimes be inconvenient if we have new formats supported. For instance, when QOI.jl is introduced in ImageIO, all ImageIO users automatically get QOI supports.

> Should I still have https://github.com/KristofferC/QOI.jl/blob/577645ab3e5930231f4043543d1c622824ceca48/src/QOI.jl#L364-L365 in the package? Should this PR be merged in addition to the ImageIO one, or only ImageIO?

It depends on how you plan to use or advertise QOI.jl:

- If you like the unified ImageIO way then there's no need to register QOI here.
- If you want people to have `load("img.qoi")` to work with only `]add FileIO QOI`, then you can keep registering QOI here.

johnnychen94, line 493:
```suggestion
add_format(format"QOI", "qoif", ".qoi", [:QOI => UUID("4b34888f-f399-49d4-9bb3-47ed5cae4e65")], [idImageIO])
```
https://github.com/JuliaRegistries/General/pull/51497

I'll let you decide whether you want to register QOI.jl here and then we can merge.

timholy:
If I'm thinking about this correctly, the issue of "merge here or via ImageIO" isn't something that we have to decide in this PR. If we want to make ImageIO the all-in-one solution, can't we just make QOI a dependency of ImageIO? So that anytime users install ImageIO they get QOI for free?

If that's right, registering it independently here gives people more freedom if they want a more stripped-down experience. I guess the one issue would be the advice that FileIO dispenses: maybe FileIO could recommend ImageIO first and QOI second?

timholy:
@KristofferC, thanks for implementing this!

KristofferC:
> If we want to make ImageIO the all-in-one solution, can't we just make QOI a dependency of ImageIO? So that anytime users install ImageIO they get QOI for free?

I think this is done in https://github.com/JuliaIO/ImageIO.jl/pull/42.

@johnnychen94, I also think it makes sense to register here so that the normal FileIO interface works as well.

timholy:
OK to tag 1.12?

KristofferC:
Good from my p.o.v (the FileIO tests in QOI.jl pass with the master branch). https://github.com/JuliaIO/FileIO.jl/pull/356 suggests the sixel format should be merged as well before 1.12.

PR #358, opened 2022-04-28
Added support for .npy and .npz files.

Pan-Maciek:
Added support for [NPZ](https://github.com/fhs/NPZ.jl/pull/50) file format.

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -15,6 +15,7 @@ const idVegaLite = :VegaLite => UUID("112f6efa-9a02-5b7d-90c0-432ed331239a")
 const idVideoIO = :VideoIO => UUID("d6d074c3-1acf-5d4c-9a43-ef38773959a2")
 const idLibSndFile = :LibSndFile => UUID("b13ce0c6-77b0-50c6-a2db-140568b8d1a5") 
 const idJpegTurbo = :JpegTurbo => UUID("b835a17e-a41a-41e7-81f0-2f016b05efe0")
+const idNPZ = :NPZ => UUID("15e1cf62-19b3-5cfa-8e77-841668bca605")
 
 ### Simple cases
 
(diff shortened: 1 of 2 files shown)


src/registry.jl
```diff
@@ -26,6 +27,8 @@ add_format(format"JLD2", (unsafe_wrap(Vector{UInt8},"Julia data file (HDF5), ver
 add_format(format"GZIP", [0x1f, 0x8b], ".gz", [:Libz => UUID("2ec943e9-cfe8-584d-b93d-64dcb6d567b7")])
 add_format(format"BSON",(),".bson", [:BSON => UUID("fbb218c0-5317-5bc6-957e-2ee96dd4b1f0")])
 add_format(format"JLSO", (), ".jlso", [:JLSO => UUID("9da8a3cd-07a3-59c0-a743-3fdc52c30d11")])
+add_format(format"NPY", "\x93NUMPY", ".npy", [idNPZ])
+add_format(format"NPZ", "", ".npz", [idNPZ])
```
johnnychen94, line 31: Is it always a gzip file here? If so might be worth using the gzip magic bytes.

johnnychen94 (approved):

src/registry.jl
Pan-Maciek: I just checked:

```python
np.savez("test.npz", a=np.array([1,2,3]), b=np.array([1.2, 3.4]))
```

Generated file looks just like a plain `.zip` file for me. 
The file magic is just zip file magic "PK\x03\x04". 
There is no comment in [central directory](https://en.wikipedia.org/wiki/ZIP_(file_format)#Central_directory_file_header) to detect.

Idk if it would make sense as it might conflict with a zip files in the future.

johnnychen94: It's okay here, don't worry. FileIO queries the file/io format in the following order:

1. check the extension `.npz`
2. check if the first few bytes match the magic bytes registered for `.npz` format.
3. if not, return `Format{:UNKNOWN}` or error.

Pan-Maciek: Ok, I added them.  
Either way I'm waiting on NPZ to approve this [MR](https://github.com/fhs/NPZ.jl/pull/50) before we can proceed. This is why I marked this as a draft.

PR #362, opened 2022-05-27
docs: explain what world-age issue is

johnnychen94:


docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+julia> save("indexed_image.png", img)
+```
+
+and **reopen a new julia REPL** and wrap the `load` into a function, you'll observe something like
+this:
+
+```julia
+julia> using FileIO
+
+julia> f() = length(load("indexed_image.png"))
```
timholy, line 51:
```suggestion
julia> f() = size(load("indexed_image.png"))
```
Some people may not quickly see the connection between `length` and the error about `size`.

docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+This is because when you first call function `f()`, the IndirectArrays package is not yet loaded and
+thus `size(::IndirectArray)` method is not defined yet. This is exactly what world age means: the
+compiled `f()` function is outdated because new methods are dynamically added in runtime. The
+perhaps surprising fact is that if you retry `f()`, it just works:
+
+```julia
+julia> f()
+16
```
timholy, line 74:
```suggestion
(4, 4)
```

docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+ERROR: MethodError: no method matching size(::IndirectArrays.IndirectArray{ColorTypes.RGB{FixedPointNumbers.N0f8}, 2, UInt8, Matrix{UInt8}, OffsetArrays.OffsetVector{ColorTypes.RGB{FixedPointNumbers.N0f8}, Vector{ColorTypes.RGB{FixedPointNumbers.N0f8}}}})
+The applicable method may be too new: running in world age 32711, while current world is 32746.
+Closest candidates are:
+  size(::IndirectArrays.IndirectArray) at ~/.julia/packages/IndirectArrays/BUQO3/src/IndirectArrays.jl:52 (method too new to be called from this world context.)
+  size(::AbstractArray{T, N}, ::Any) where {T, N} at abstractarray.jl:42
+  size(::Union{LinearAlgebra.Adjoint{T, var"#s962"}, LinearAlgebra.Transpose{T, var"#s962"}} where {T, var"#s962"<:(AbstractVector)}) at ~/packages/julias/julia-latest/share/julia/stdlib/v1.9/LinearAlgebra/src/adjtrans.jl:173
+  ...
+Stacktrace:
+ [1] length(t::IndirectArrays.IndirectArray{ColorTypes.RGB{FixedPointNumbers.N0f8}, 2, UInt8, Matrix{UInt8}, OffsetArrays.OffsetVector{ColorTypes.RGB{FixedPointNumbers.N0f8}, Vector{ColorTypes.RGB{FixedPointNumbers.N0f8}}}})
+   @ Base ./abstractarray.jl:291
```
timholy, line 64: This will need to be fixed.

docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+This is because when you first call function `f()`, the IndirectArrays package is not yet loaded and
+thus `size(::IndirectArray)` method is not defined yet. This is exactly what world age means: the
+compiled `f()` function is outdated because new methods are dynamically added in runtime. The
+perhaps surprising fact is that if you retry `f()`, it just works:
```
timholy, line 70:
```suggestion
To understand why this happened, you have to understand the order of events:

1. When calling `f()` from the REPL, Julia first compiled `f`. Importantly, when compiling, Julia didn't know what type of object was going to be returned by `load`, so in the compiled code it waits to see what object actually gets returned before figuring out which method of `size` to call. (This is called *runtime dispatch*.)
2. It queried the file, recognized a PNG file, and *loaded the ImageIO and PNGFiles packages*. (It's for the loading of these packages that you needed to start a fresh Julia session.)
3. FileIO calls the appropriate PNG-specific `load` function in PNGFiles. (We'll have more to say about this step further below.) This causes an image to be returned, which is an array of a type defined by the IndirectArrays package (a dependency of PNGFiles). 
4. `f` calls `size` on the returned image. However, this fails, because at the time you called `f`, the IndirectArrays package wasn't loaded.

In other words, `size` method lives in a world that's newer than the one from which you called `f()`.
This leads to the observed error.

!!! note
    World age is crucial to Julia's ability to allow
    you to *redefine* methods interactively, but the error we're illustrating is an unfortunate side-effect.

The good news is it's easy to fix, just try calling `f()` again:
```

docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+
+```julia
+julia> f()
+16
+```
+
+The second `f()` works because this time you're running a recompiled version of `f()` in the latest
+world age with the necessary `size(::IndirectArray)` defined.
```
timholy, line 78:
```suggestion
The second `f()` works because this time you're calling `f()` in the latest
world age with the necessary `size(::IndirectArray)` already defined.
In essence, you fast-forward to the latest world with each statement you
type at the REPL.
```

docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Solution
```
timholy, line 80:
```suggestion
## Solutions

### `Base.invokelatest`

One solution is to make the call to `size` via `Base.invokelatest`,
which exists explicitly to work around this world-age dispatch problem. Literally, `invokelatest`
dispatches the supplied call using the latest world age (which may be newer than when you
typed `f()` at the REPL). **In a fresh Julia session**,

``julia
julia> using FileIO

julia> f() = Base.invokelatest(size, load("indexed_image.png"))
f (generic function with 1 method)

julia> f()
(4, 4)
``

!!! note
    In step 3 above ("FileIO calls the appropriate PNG-specific `load` function in PNGFiles"),
    the call to the `load` function defined in PNGFiles is made via `invokelatest`.
    Otherwise, even ordinary interactive usage of FileIO (without burying `load` inside a function)
    would cause world-age errors.

!!! warning
    Using `invokelatest` slows your code considerably. Use it only when absolutely necessary.

### Eagerly load the required packages first
```

timholy: You're going to have to change double-backticks to triple-backticks; unfortunately GitHub terminates the suggestion on the first triple-backticks it finds.

docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+
+The solution to the world age issue is simple: **eagerly load the needed packages**. For instance,
```
timholy, line 82:
```suggestion
Another solution to the world age issue is simple and doesn't have long-term downsides:
**eagerly load the needed packages**. For instance,
```

docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+if you're seeing world age issue complaining methods related to `IndirectArray`, then load
+IndirectArrays eagerly:
+
+```julia
+julia> using FileIO, IndirectArrays # try this on a new Julia REPL
+
+julia> f() = length(load("indexed_image.png"))
```
timholy, line 89:
```suggestion
julia> f() = size(load("indexed_image.png"))
```

docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+f (generic function with 1 method)
+
+julia> f()
+16
```
timholy, line 93:
```suggestion
(4, 4)
```

docs/src/world_age_issue.md
```diff
@@ -0,0 +1,109 @@
...
+
+# This ensures that whoever loads `MyFancyPackage`, he has IndirectArrays loaded and
+# thus avoid the world age issue.
+using IndirectArrays, FileIO
+
+f(file) = length(load(file))
+end
+```
+
+Enjoy the FileIO and its lazy loading, but be aware that this isn't the silver bullet for latency.
```
timholy, line 109:
```suggestion
Enjoy the FileIO and its lazy loading, but be aware that its speedy loading comes with some caveats.
```

timholy (commented):
Sorry for the "long latency" :wink:  here! And thanks for doing this.

docs/src/world_age_issue.md
johnnychen94: A tip is to wrap the block with four backticks

image

johnnychen94:
@timholy thanks for the heavy rewrite, much clearer!

PR #363, opened 2022-07-06
Add Gziped fits file support and AstroImages.jl library

sefffal:
This PR adds support for loading GZiped FITS files.
Astronomical FITS files are often gziped and given an extension like `.fits.gz`. Compressed fits files can be loaded directly by FITSIO.

I did my best to untangle the file loading between plain GZiped files, GZiped RData files, and GZiped fits files. 
Everything should work exactly as before unless a filename matches the `.fits.gz` pattern.
I believe this also fixes one `#FIXME` comment in `tests/query.jl` for compressed RData.

Finally, I also add support for the AstroImages library as an alternative to FITSIO which was my original purpose 😅.

### DIFF
--- a/src/registry.jl
+++ b/src/registry.jl
@@ -24,7 +24,6 @@ add_format(format"JLD", (unsafe_wrap(Vector{UInt8}, "Julia data file (HDF5), ver
                          unsafe_wrap(Vector{UInt8}, "Julia data file (HDF5), version 0.1")), ".jld", [:JLD => UUID("4138dd39-2aa7-5051-a626-17a0bb65d9c8")])
 add_format(format"JLD2", (unsafe_wrap(Vector{UInt8},"Julia data file (HDF5), version 0.2"),
                           unsafe_wrap(Vector{UInt8}, "HDF5-based Julia Data Format, version ")), ".jld2", [:JLD2 => UUID("033835bb-8acc-5ee8-8aae-3f567f8a3819")])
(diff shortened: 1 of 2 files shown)


sefffal:
Three-week bump.

src/registry.jl
```diff
@@ -60,6 +59,9 @@ end
 
 detect_compressed(io, len=getlength(io); kwargs...) = detect_compressor(io, len; kwargs...) !== nothing
 
+const compressed_fits_exten = r"\.(fit|fits|fts|FIT|FITS|FTS)\.(gz|GZ)\>"
+name_matches_compressed_fits(io) = (:name ∈ propertynames(io)) && endswith(io.name, compressed_fits_exten)
```
johnnychen94, line 63: Curiously, what if `propertynames` doesn't contain `:name`? Will this break the functionality of `detect_rdata`?

How about

```julia
name_matches_compressed_fits(io::IOStream) = endswith(io.name, compressed_fits_exten)
name_matches_compressed_fits(io::IO) = false # does FITSIO and AstroImages support generic IO object (e.g., `IOBuffer`)?
```

johnnychen94 (approved):
cc: @jishnub @mileslucas

mileslucas (approved):
I'm unexperienced with the Rdata streams, but the astroimages looks good to me- I've tested locally with success with both FITSIO.jl and AstroImages.jl

src/registry.jl
sefffal: Looking at the logic of this again, I think it's okay. 
If we detect that it's compressed, then if there's no name or the name does not match a compressed fits file, it falls back to the compressed R data path.

Ideally someone with compressed R data could remove the file extension and test this path out.

PR #365, opened 2022-09-08
Add .bib files to FileIO interface

drelatgithub:
The .bib files store bibliography data which can be used in a LaTeX document. This PR registers ".bib" files with [Bibliography.jl](https://github.com/Humans-of-Julia/Bibliography.jl), which can handle parsing / importing / exporting / processing the Bibliography data.

The FileIO interfaces have been added in https://github.com/Humans-of-Julia/Bibliography.jl/pull/47


drelatgithub:
I think that this PR would benefit communities that deal with bibtex management. Please let me know if any help is needed in the reviewing process!

johnnychen94 (approved):

PR #366, opened 2022-11-23
Preemptively fix tests that are about to break

apaz-cli:
Apply these changes if https://github.com/JuliaLang/julia/pull/47369 gets merged.

The error message for MethodError currently applies to both calls to functions and `invoke()`, but did not specify which the error came from. Fixing this will break some tests, hence this PR.


ViralBShah:
@apaz-cli Would you have the cycles to help update the tests here? Happy to give you commit access to these repos.

ViralBShah:
Tests are working with #398

PR #367, opened 2022-12-02
Add Arrow file format

mortenpi:
Also known as [Feather v2](https://arrow.apache.org/docs/python/feather.html#feather-file-format).

* Official spec, for magic bytes (`ARROW1`, padded to 8-bytes with zeros): https://arrow.apache.org/docs/format/Columnar.html#ipc-file-format
* IANA assignment (magic bytes; extension): https://www.iana.org/assignments/media-types/application/vnd.apache.arrow.file
* Julia package: https://github.com/apache/arrow-julia


sjkelly:
I don't see any other packages that have this functionality, so this seems reasonable.

PR #384, opened 2024-02-27
Remove QuartzImageIO

ViralBShah:
As discussed in https://github.com/JuliaIO/QuartzImageIO.jl/issues/77


ViralBShah:
Failures on nightly are not related to this PR. I think we should be able to merge.

kimikage:
The registry table on documentation is not updated. https://juliaio.github.io/FileIO.jl/dev/registry/#Registry-table
Maybe it would be better to generate the table automatically or add something like doctest.

PR #385, opened 2024-02-27
Update CI for mac aarch64, codecov 4 token

ViralBShah:
I wonder if we can drop testing for 1.0 and save CI time.


ViralBShah:
Wonder why the tests didn't run on all the architectures.

PR #388, opened 2024-03-06
Update registry.jl

Larbino1:
Change to stl binary/ascii type detection. Removed check on value of final attr bits, which caused a failure to parse for me, incorrectly treating it as an ascii type stl. If the eof is in the correct place, this value does not matter.

@SimonDanisch  Maybe as a MeshIO contributor you can give this a quick look? Its a tiny change!


SimonDanisch:
Could you add a small test file?

Larbino1:
> Could you add a small test file?

Yeah sure. 

Do you mean:
- A test in FileIO
- Checking that detect_stlascii, and detect_stlbinary work
- By adding to the repo an STL of each type, and testing both

SimonDanisch:
Add a test file here: https://github.com/JuliaIO/FileIO.jl/blob/d30fbd7d2edb1c234f4a361967e1b063b9dea447/test/query.jl#L356-L366

Larbino1:
@SimonDanisch 

I've included the file as a test case. Please let me know if you would like any other work, or if you can support merging.

I guess eventually supporting STLs with attribute data would be nice, but for this specific problem (a binary stl, with nonzero 'attribute byte count' but actually containing no attribute byte data), things work correctly rather than being misconstrued as an ASCII stl.

Larbino1:
Just bumping this again @SimonDanisch @sjkelly 

It works, the test is in, the change non-breaking. If there is anything I can do to help get this merged let me know.

PR #390, opened 2024-06-04
Reduce amount of CI

ViralBShah:


ViralBShah:
The Documenter.jl generation needs to be updated.

PR #392, opened 2024-07-10
Fix syntax in example in documentation

PatrickHaecker:
A parenthesis was missing.


nhz2:
Thank you for fixing this.

PR #393, opened 2024-07-13
Documentation site update

abhro:
- Use Documenter.jl v1
- Clean installation instructions
- Add headers for docstrings
- Use `Downloads.download` instead of `Base.download`
- Add cross-reference links


src/FileIO.jl
```diff
@@ -11,6 +11,7 @@ export DataFormat,
        add_saver,
        filename,
        file_extension,
+       info,
```
nhz2, line 14: Why is `info` being exported in this PR?

abhro: `info` is already listed in the `FileIO` module docstring, so I added a link and the docstring to the documentation site. The export is just to make more sure that it's part of the public API.

nhz2: Other people may have opinions on exporting `info` so it would be good to make that change in a separate PR.

abhro: That's fair. I'll push a revert soon. Thanks!

abhro: Reverted in [a33d30d](https://github.com/JuliaIO/FileIO.jl/pull/393/commits/a33d30df84639e3fd312afc0a2e426b9a1e2d732)!

nhz2:
Thank you for fixing this up

PR #397, opened 2024-08-13
Make Formatted <: IO to avoid invalidations

serenity4:
Addresses #396, and further seems to simplify the implementation a little.

This change is a bit weird, as `Formatted` itself does not semantically represent an `IO` type, it is only `Stream` that does. But since `Stream` can only subtype a single abstract type, that is the only way to make it `Stream <: ... <: IO` such that the type hierarchy is preserved.

The potential drawback I see is that it probably causes more specialization, threading `Stream` through IO functions. At first I attempted to keep the more specific `read`/`write` methods for `Stream`, but because `Stream` is now an `IO` ambiguity errors pop up which require more methods to be defined manually to be resolved, at which point I thought having a minimal IO implementation might be better. I can try to maximally avoid specialization by adding these methods back, but it is quite likely that other ambiguity errors pop up in the future if going down that path. I believe that the invalidation fixes outweigh the added specialization.

I'm happy to discuss/drop this PR if the subtyping change it not desired, in which case we can try to come up with another solution to #396 (if it exists). I'm also curious to know if there was some discussion about having `Stream <: IO`, seeing that it implements lots of IO interface functions.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -21,7 +21,7 @@ end
 formatname(::Type{DataFormat{sym}}) where sym = sym
 
 
-abstract type Formatted{F<:DataFormat} end  # A specific file or stream
+abstract type Formatted{F<:DataFormat} <: IO end  # A specific file or stream
 
 formatname(::Formatted{F}) where F<:DataFormat = formatname(F)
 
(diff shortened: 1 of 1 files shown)


serenity4:
The discussion at https://github.com/KristofferC/OhMyREPL.jl/issues/355 makes me believe that it isn't really possible, unfortunately (see notably [this comment](https://github.com/KristofferC/OhMyREPL.jl/issues/355#issuecomment-2269939030)).

nhz2:
My understanding of that comment makes me think that there are a few options to improve things:
1. Don't have `AnnotatedString` subtype `AbstractString`.
2. Have `AnnotatedString` subtype `AbstractString`, but existing `AbstractString` functions should ignore annotations.

serenity4:
I believe in theory these were both considered, as explained in the answer to that comment: https://github.com/KristofferC/OhMyREPL.jl/issues/355#issuecomment-2270341118. In practice, it seems like the slightly less intuitive/ideal design brings good trade-offs, with less code churn and good behavior, one downside being invalidations popping up. Ideally it's not downstream code that should deal with this, but I'm exploring whether we could remove this deliberate downside without noticeable impact here.

serenity4:
As `Stream` fundamentally represents some IO-like type, would it be possible to slightly change the design such that `Formatted` is more of a trait, and does not require to bite the subtyping silver bullet? This way, we could have `Stream` subtype `IO` directly. What do you think?

nhz2:
There is already a `stream(s::Stream)` function to get an `IO` representation of a `Stream`, so I'm not sure subtyping `IO` is needed.

nhz2:
This should be fixed by https://github.com/JuliaLang/julia/pull/55593

I think it might make sense for `Stream` to be an `IO`, but I don't think `File` should be an `IO`

Also making `Stream` an `IO` will require implementing or forwarding most of the `IO` methods. It is not enough to just forward `read` and `write` for single bytes.

serenity4:
Nice, thanks!

PR #401, opened 2024-11-02
CI: Expanded testing to include julia LTS versions

stemann:
Expanded testing to include both current, and previous LTS versions, in addition to the current stable version - on all 64-bit tier 1 CPU archs. Also added macOS (not just macOS on aarch64 using current stable version).

Also, fixed invalidations by using julia 1.10.


stemann:
Also, fixes test target for (ubuntu-latest, x86, 1), as the arch was not passed to setup-julia - so tests were not running on a 32-bit Julia.

nhz2:
Thank you for fixing this.

PR #403, opened 2024-11-13
Introduce `HTTPExt` package extension

Drvi:
So the package is easier to statically precompile on newer Julia versions


Keno (approved):
LGTM, but I'm not a regular contributor, so I'll leave it to someone else

PR #407, opened 2024-11-20
Adjust to Julia 1.12 change to implicit world age increment

Keno:
Julia 1.12 will no longer implicitly increment world ages within expressions. However, we will likely retain world age increments between statements of a `@testset`, so all that's needed here is to move the `load` to a separate like. More explicitly, the `collect`s could be `invokelatest`'ed instead. See https://github.com/JuliaLang/julia/pull/56509


Drvi (approved):
LGTM!

Linking this comment as it directly addresses this package: https://github.com/JuliaLang/julia/pull/56509#issuecomment-2488280353

